PR #4, opened 2022-03-29
feat(docker): dockerized ui

valearna:
- created dockerfile with separate stages for build and deployment
- deployment based on nginx
- added docker-compose configuration for easy deployment (docker-compose up)
- added PORT variable to .env to set deployment port


azurebrd (approved):
That's great, thanks @valearna .  The .env, docker-compose.yml, and Dockerfile all make sense.  The nginx.conf I don't know much about, is that a standard deploy ?  How do we deploy literature and stage-literature differently, change the .env PORT ?

azurebrd:
Cool cool

PR #5, opened 2022-04-29
faceted search

valearna:
Added faceted search on pubmed_types


azurebrd (approved):
That's great, thanks for pre-explaining it @valearna :)

PR #6, opened 2022-05-02
Scrum 1489 - faceted search

valearna:
Fixed layout on large displays


azurebrd (approved):
Cool cool, I expected you'd get rid of the container eventually, given your preferences against rows and columns :)

PR #8, opened 2022-05-13
Scrum 1502 spinners + improved layout

valearna:
- Added spinners inside search buttons for Query and Sort pages
- Removed white div under search box in Query page
- Added 'no results found' for both pages


azurebrd (approved):

PR #9, opened 2022-05-13
Scrum 1461 bread crumbs

valearna:
- Added bread crumbs for facets
- changed logic for adding/removing facets values


azurebrd (approved):
Looks good ! @valearna

PR #10, opened 2022-05-17
Scrum 1462 - Added more facets

valearna:
- Added two more facets to the "bibliographic data" category
- Created separate component for show more/less/all buttons
- Changed logic to manage show more/less/all buttons: hide them when no all entries already displayed
- Fixed breadcrumbs: they were displayed on separate lines for each facet, now on the same line


azurebrd (approved):

PR #12, opened 2022-05-27
Search fixes

valearna:
This PR fixes the following issues:

- warning about undefined value for searchQuery text
- if you go to the search page, query for a PMID and come back, the search results load as if someone had searched for an empy string
- excluded warnings on useEffect in Facets.js


azurebrd (approved):
Thanks @valearna !

PR #13, opened 2022-07-12
Jenkins test

valearna:
Removed nginx config file from docker-compose volumes as it was causing issues with jenkins. I had to set up jenkins by using docker commands directly and not docker-compose, at least for now. Merging this PR into main should also trigger a new build


azurebrd (commented):
Sorry, I missed that

PR #14, opened 2022-07-15
Scrum 1357 added facets for "in corpus" and "needs review"

valearna:
- Added facets to new facet group "Alliance Metadata"
- fixed several UI issues, including re-running search after all facets are removed from the search criteria, sorting facets to display by the array defined in the UI and not by the list returned by ES (which could be arbitrary), added facet types to breadcrumbs to differentiate between criteria when the facet value is the same (e.g., for mod in corpus and mod needs review), and using separate accordions to avoid closing one facet group when another one is opened


azurebrd (approved):
Thanks @valearna !

PR #15, opened 2022-07-20
Scrum 1357 - open facet groups with selected facet values when reopening search page

valearna:


azurebrd (approved):
Thanks @valearna .  I can't really picture what it does, but the code looks good.

PR #17, opened 2022-08-29
SCRUM-1770 Adding Extra Fields to search results

pjhale:
Added xRefs, Authors, pub date, and abstract.  Needs the corresponding 
pr in the API to have access to this data.  Adds red obsolete to 
obsolete xrefs and cuts abstracts down to 500 chars.

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -248,7 +248,7 @@
 
 .Col-general {
   text-align: left;
-  display: block;
+  display: flex;
   align-items : center;
   color: var(--font-black);
   border: 2px solid white;
@@ -265,7 +265,7 @@
 .Col-display-left {
   border-radius: 30px 0 0 30px;
 }
-.searchRow {
+.Col-search {
   display: block;
 }
 .searchRow-title{
--- a/src/App.css
+++ b/src/App.css
@@ -248,7 +248,7 @@
 
 .Col-general {
   text-align: left;
-  display: flex;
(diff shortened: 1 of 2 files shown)


azurebrd (commented):
Hi @pjhale .  The .Col-general is used a lot in the Biblio section.  If I could see how that looks and it generally looks the same, then display: block sounds good.  But if it doesn't, we might want a different css entry for the Search page to use

pjhale:
Changed it so it has its own class for the block style... everything else from col-general looks good there.

azurebrd (approved):
Cool, if that works for you, that's great @pjhale 
I don't recall how .css inheritance works if there's 2 classes that set the same field to different values (Col-general display-> flex, Col-search display->block, new Col in line 28 with both classNames)  I think for equal specificity, what takes precedence is what order they show up in the .css file with the later values taking precedence, but if you know for sure, let me know :)

azurebrd:
@pjhale From the comment about the API PR, I'm not sure if this is ready to merge.  I think the moment it merges, @valearna set it so that it will update stage-literature UI to run that code, so if it's dependent on the API PR we could just leave this ready to Merge.

pjhale:
This needs the following API PR to work properly...  let me know if there is a better way to link these PR's....

https://github.com/alliance-genome/agr_literature_service/pull/222

PR #18, opened 2022-09-23
feat(search): added new 'ID' search field

valearna:
- unified search interface for xrefs and title searches with new search field 'ID'
- when searching for an AGR curie (starting with AGR: or AGRKB:) clicking the search button automatically redirects to the biblio edit page for the inserted curie
- other IDs are managed using functions from the old xref query form and if the xref is found the page redirects to the biblio edit with the respective agr curie

### DIFF
--- a/src/actions/searchActions.js
+++ b/src/actions/searchActions.js
@@ -48,23 +48,36 @@ export const fetchInitialFacets = (facetsLimits) => {
   }
 }
 
-export const searchReferences = (query, facetsValues, facetsLimits, sizeResultsCount) => {
+export const searchReferences = (query, fieldToSearch, facetsValues, facetsLimits, sizeResultsCount) => {
   return dispatch => {
-    dispatch(setSearchLoading());
-    dispatch(setSearchQuery(query));
-    dispatch(setSearchFacetsValues(facetsValues));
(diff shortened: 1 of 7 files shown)


valearna:
@pjhale I've unified the search interfaces but kept the old functions for searching xrefs. I've added a special case for AGR curies that redirects directly to the biblio edit page. After talking to Ceri I realized that there's no need to use elasticsearch for these searches as we don't need to search for obsolete xrefs or curies

azurebrd (approved):
That's great, thanks @valearna 

In components/search/Facets.js line 155-7 I understand that it makes sense, but it's a bit confusing without { } or indentation that 156 is the then part, or just removing the if in 155 and making an && to connect to line 156.  It's find though.

searchActions line 75-79 indentation is the same as before, so it's fine, but it's also a bit off.

valearna:
Thanks @azurebrd! 

I've added {} and fixed indentation in Facets.js for lined 155-7.

In searchActions the indentation looks off, but it's what pycharm is applying for standard js code, so I'd keep it as is

azurebrd (approved):
Thanks @valearna !

Oh, interesting about how pycharm does it.  Good to know :)

PR #19, opened 2022-09-23
Scrum 1769 (second version) - query id to open reference directly from biblio edit page

valearna:


azurebrd (approved):
Great, thanks @valearna !

azurebrd:
response_found in biblioActions.js wasn't used anymore, after merging I pushed a change removing that

azurebrd:
@valearna 
now that biblioReducer sets         getReferenceCurieFlag: true   we might not need SET_GET_REFERENCE_CURIE_FLAG at all, since that's just called from Sort.js and Create.js to always set it to true in conjunction with setReferenceCurie(reference['curie'])   So probably we can get rid of that action in both components.  I'm not sure why they needed to be set independently before, maybe it was a mistake, or maybe it was before of how setting values across reducers was confusing.

PR #20, opened 2022-10-03
SCRUM-1356 Adding Pagination

pjhale:
Adding code to delivery pagination.  Functions are limited to First,  Last, Next, and Previous.  The first page is defined as page zero (to  make the maths easier) but displayed as page one.  Page is reset to zero  upon changing the size of pages or running another query.  Its possible that all the pagination could be moved into its own component as it get more complex... or now if we think thats necessary.

Needs the corresponding SCRUM-1356 pull from the agr_literature_service repo to function properly.

### DIFF
--- a/src/actions/searchActions.js
+++ b/src/actions/searchActions.js
@@ -1,6 +1,7 @@
 import axios from "axios";
 
 export const SEARCH_SET_SEARCH_RESULTS_COUNT = 'SEARCH_SET_SEARCH_RESULTS_COUNT';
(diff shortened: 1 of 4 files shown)


azurebrd (approved):
Cool, thanks @pjhale !

valearna:
I've checked out the branch and the search page still shows the "query by exact xref" component at the top. Is this branch synced with main?

valearna:
the new pagination component looks good, thanks @pjhale

valearna (approved):

pjhale:
I haven't pulled in main since I branched but no merge issues seem to exist.

azurebrd:
@pjhale Probably best to merge main into your branch, see if it seems okay, and then update this PR with that

PR #21, opened 2022-10-04
fix(search): allow search by facets only

valearna:
Selecting a facet without any text in the search bar now triggers a search


azurebrd (approved):
Thanks @valearna !
Thanks for the { } for indentation also, but I was okay with it the other way too :P

PR #23, opened 2022-10-10
Xref URL

pjhale:
Adding the ability to click on an Xref direct from the search results  page.  Uses the searchXref method to call the cross_references endpoint. This will call every xref in a different api call but it doesn't seem to cause any problems.  Tested with a search artificially expanded to 500 records with no errors.

Of note... MGI urls seem to be a blank informatics.jax.org.  I think this is a known issue but noting it here all the same.

### DIFF
--- a/src/components/search/SearchResults.js
+++ b/src/components/search/SearchResults.js
@@ -13,9 +13,9 @@ const XrefElement = (xref) => {
     const [url, setUrl] = useState(null);
     useEffect(() => {
       searchXref(xref.xref.curie, setUrl);
-    }, [url]); // eslint-disable-line react-hooks/exhaustive-deps
+    }, []); // eslint-disable-line react-hooks/exhaustive-deps
     return (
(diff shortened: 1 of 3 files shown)


azurebrd (approved):
Great, thanks @pjhale !

src/components/search/SearchResults.js
```diff
@@ -1,14 +1,23 @@
...
 import {Modal} from 'react-bootstrap';
-import {setSearchError} from '../../actions/searchActions';
+import {setSearchError, searchXref} from '../../actions/searchActions';
 import Button from 'react-bootstrap/Button';
 
+const XrefElement = (xref) => {
+    const [url, setUrl] = useState(null);
+    useEffect(() => {
+      searchXref(xref.xref.curie, setUrl);
+    }, [url]); // eslint-disable-line react-hooks/exhaustive-deps
```
valearna, line 16: searchXref modifies url and we want to execute the useEffect only once, so url should be removed from the useEffect deps

valearna (approved):

valearna:
thanks @pjhale!

PR #24, opened 2022-10-12
Correcting URL's

pjhale:
It seems that this endpoint returns URL in two different fields.  The  pages field that is an array and the url field on the top level.  This  change means if the pages field exists we use the first url from that  array and if not we use the top level url.  I dont know if its possible  to have more than one entry in this array but the returns I looked at  did not seem to.  Checked the examples that Ceri provided (MGI, SGD, etc) and all appear to be going to the correct url now.

### DIFF
--- a/src/actions/searchActions.js
+++ b/src/actions/searchActions.js
@@ -67,7 +67,11 @@ export const searchReferences = (query, facetsValues, facetsLimits, sizeResultsC
 export const searchXref = (xref, setUrl) => {
   axios.get(restUrl + '/cross_reference/'+xref)
   .then(res => {
-    setUrl(res.data.url);
+    if(res.data.pages){
+      setUrl(res.data.pages[0].url)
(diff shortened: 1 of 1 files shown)


azurebrd (approved):
@pjhale The reference links from the mods should have a 'pages' attribute based on the https://github.com/alliance-genome/agr_schemas/blob/master/resourceDescriptors.yaml but PMIDs and DOIs wouldn't.   ZFIN resources could have multiple pages, but I don't think any references could have multiple pages.  (Maybe other things could also have multiple pages).  I don't know how the API should work, Adam wrote it.  Maybe @valearna has an opinion on whether it should always resolve down to a single URL (I don't know)

valearna:
I don't know how the API should work and why it was designed like this - we can talk about it, but for now this PR looks good

valearna (approved):

PR #26, opened 2022-10-24
Scrum 2137

azurebrd:


azurebrd:
Hey @valearna .  This has the changes to make the date_published + start/end work in the UI and update through the API.  If this looks good, I'm probably going to break up the Biblio component into separate components for Entity, Workflow, Display, Editor.

valearna (approved):
Looks good, thanks @azurebrd. The published date field is not present in the display view but only in the biblio edit. Is this the intended behavior?

PR #29, opened 2022-11-18
Author Filter Tweaks

pjhale:
Moving author filter to its own component, moving that component to the  top of the author box, adding a clear author filter button, removing  authorFilter from text box after selecting author.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
     "react-bootstrap": "^1.6.0",
     "react-dom": "^17.0.2",
     "react-icons": "^4.3.1",
+    "react-loading-overlay": "^1.0.1",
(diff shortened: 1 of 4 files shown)


valearna:
@pjhale I've added an overlay to make it easier to understand when facets are loading after changing the limits or the author filter. Let me know if you like it, otherwise feel free to git reset to the previous commit

pjhale:
I think this looks really good... and it feels better using the author filter now that the whole page doesn't reload... only the relevant information.  Couldn't find any bugs with the new events either.  I'm ready to put in this PR if you are.

valearna:
Thanks @pjhale

PR #33, opened 2022-12-07
feat(tarball): Added link to download tarball with additional files

valearna:


azurebrd (approved):
Great, thanks @valearna !
(It's kind of confusing how long the ternary statement is, and that it's embedded inside another ternary statement, but it makes sense, so I think it's fine)

valearna:
Thanks @azurebrd, I tried to reformat the code for that component and everything seems much more readable. Let me know if this is better

azurebrd:
Great, thanks @valearna !

PR #36, opened 2023-01-23
Scrum-1464 Adding Search Types

pjhale:
Added the ability to search via different fields.  Will need the API pull to function correctly.  Also did a clean up of the search references function.  This function does not require any arguments anymore but draws directly from the redux store.  Leaving this as draft until we can test with the new analyzer/fields in ES.

### DIFF
--- a/src/components/search/Facets.js
+++ b/src/components/search/Facets.js
@@ -42,36 +42,37 @@ LoadingOverlay.propTypes = undefined;
 
 const DateFacet = ({facetsToInclude}) => {
 
-  const searchState = useSelector(state => state.search);
+  const datePubmedModified = useSelector(state => state.search.datePubmedModified);
+  const datePubmedAdded = useSelector(state => state.search.datePubmedAdded);
   const dispatch = useDispatch();
 
   return (
     <div>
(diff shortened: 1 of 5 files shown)


src/components/search/Facets.js
```diff
@@ -42,36 +42,37 @@ LoadingOverlay.propTypes = undefined;
...
 
   return (
     <div>
       <div key={facetsToInclude[0]} style={{textAlign: "left", paddingLeft: "2em"}}>
       <h5>{facetsToInclude[0]}</h5>
-        <DateRangePicker value={searchState.datePubmedModified} onChange= {(newDateRangeArr) => {
+        <DateRangePicker value={datePubmedModified} onChange= {(newDateRangeArr) => {
           if (newDateRangeArr === null) {
             dispatch(setDatePubmedModified(''));
             dispatch(setSearchResultsPage(0));
```
valearna, line 56: the second and the third dispatch calls are always the same for both cases - they can be placed outside the if-else. Same thing for lines 70-71 and 74-75

valearna (approved):
Looks good, thanks @pjhale

src/components/search/Facets.js
pjhale: Moved these actions.

PR #39, opened 2023-02-06
Scrum 2474 - added pagination to topic and entity tags page

valearna:
- using new api endpoints to get the list of tags and the curie->name mappings
- removed data and fetch functions from the redux actions/reducers - now managed in the component
- copied pagination component from search
- set default page size to 10


azurebrd (approved):
Great, thanks @valearna !

PR #42, opened 2023-02-08
Scrum 2473 sort by entity type

valearna:
- added icon to sort by entity type
- icon changes color when clicked to indicate that the table is being sorted by entity type
- multiple clicks alternate between asc and desc sorting
- modified api call to include new sort by parameters


azurebrd (approved):
Thanks @valearna !

PR #43, opened 2023-02-09
fix Sort label scrum-2552, result show number scrum-2550

FB-ping:


pjhale:
Removed some old code that wasn't doing anything.  Other than that I think its fine.

pjhale (approved):

PR #44, opened 2023-02-13
feat(topic_entity_tag): new icons for sorting

valearna:
Clicking three times goes back to non-sorted results


azurebrd (approved):
Cool !  I think it does a 3 way toggle between not / down / up ?  But there's only an up and down icon, and not a sort icon for the not state ?

valearna:
Yes, I couldn't find the double icon arrow and I set the down arrow (not colored) as default, but I can add the default sort icon if we find it

azurebrd:
Gotcha, cool

valearna:
I found the icons in font awesome, but not in react icons (which is supposed to be a meta project containing also fa icons). I'll update them

PR #46, opened 2023-02-17
scrum-2100 fix html for italics

FB-ping:


azurebrd (commented):
Looks good to me, but I'll defer to @valearna and @pjhale

pjhale:
Looks good to me... I was able to get authors in both areas with non-standard characters and they displayed correctly.

valearna (approved):

PR #48, opened 2023-02-21
Fix abstract truncation

pjhale:
Only split on whitespace.  Should fix html characters only getting part of their codes.


valearna (approved):
thanks @pjhale and @FB-ping

PR #51, opened 2023-02-28
feat(referencefiles): using new separate endpoint to get list of refe…

valearna:
…rence files + metadata now editable from file management tab


azurebrd (approved):
Thanks @valearna !

PR #53, opened 2023-03-03
Adding format date range function

pjhale:
Formats the date to YYYY-MM-DD for passing to ES

Prevents a different date from being passed to the API via timestamp issues.


valearna (approved):
I'm not sure why the new function is needed but everything looks good. Is this to prevent possible issues in the future?

PR #54, opened 2023-03-07
Scrum 2611 - UI using 'page' and 'page_size' instead of 'offset' and 'limit'

valearna:


src/actions/biblioActions.js
```diff
@@ -10,8 +10,8 @@ const restUrl = process.env.REACT_APP_RESTAPI;
 // const port = 11223;
 // const port = 49161;
 
-const ateamApiBaseUrl = 'https://beta-curation.alliancegenome.org/';
-const restUrl = process.env.REACT_APP_ATEAM_API_BASE_URL;
+//const ateamApiBaseUrl = 'https://beta-curation.alliancegenome.org/';
+const ateamApiBaseUrl = process.env.REACT_APP_ATEAM_API_BASE_URL;
```
azurebrd, line 14: That's weird.  I thought we'd already done this.  It's good to get this in, I'm just confused because I thought we added those .env variables and changed this .js last week

azurebrd (approved):
Thanks @valearna .  Made a separate comment about that variable not being already changed

src/actions/biblioActions.js
valearna: yes, it is weird, and I don't know how the UI could start with that double variable declaration

PR #55, opened 2023-03-15
Scrum 2701 - grant access to all files if reference is open_access

valearna:
Added check for open_access in biblio component for referencefiles and in file management tab


sweng66 (approved):
Looks good! Thanks @valearna

azurebrd (approved):
Great !

PR #56, opened 2023-03-20
SCRUM-2227 Adding in delete functionality

pjhale:
Using API delete function as it was already written so no API changes.

https://agr-jira.atlassian.net/browse/SCRUM-2227


valearna (approved):
Looks good, thanks @pjhale!

PR #57, opened 2023-03-20
SCRUM-2227 Adding check for hasAccess

pjhale:
If we needed these enforced rigidly this kind of check should be done at  the API level but this will prevent users from accidentally deleting  files they dont have access to.


valearna:
Ceri and Kimberly signed off on allowing deletion by everybody with access to the files, so your changes are good, thanks @pjhale

valearna (approved):

PR #59, opened 2023-04-13
File management page

pjhale:
Will need merge to main since the db has changed and api calls are  failing.

### DIFF
--- a/src/components/Tracker.js
+++ b/src/components/Tracker.js
@@ -105,6 +105,8 @@ const Tracker = () => {
 
   return (
     <div>
+    <h4>Reference File Tracker</h4>
+    <br/>
       { missingFileResults ?
       //<Button variant="primary" className="download-tracker-button">Download All</Button>
       <Table bordered size="sm">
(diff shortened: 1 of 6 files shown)


src/components/Tracker.js
```diff
@@ -0,0 +1,141 @@
...
+  return dispatch => {
+    axios.get(restUrl + '/reference/missing_files/'+mod_abbreviation)
+        .then(res => {
+          dispatch(setMissingFileResults(res.data));
+        })
+        .catch();
+    }
+}
+
+function getOktaModAccess(oktaGroups) {
```
azurebrd, line 23: This should probably just import the other module that exists, so we don't have to maintain 2 copies when things change ?  We're probably going to be changing something for okta Tester group, which would override access with whatever they set manually through the UI.

azurebrd (changes requested):
Looks great @pjhale .  Just the request to import the existing getOktaModAccess if they're meant to do the same thing.  But if you think it's better to have a separate one because it's a different use-case, we can talk with Valerio when he's back.  Thanks !

azurebrd (approved):
@pjhale explained that we need to _not_ have developer, so can't use the other function.  @valearna agrees it's fine to have separate functions for now until we can add some Person-based API thing or a Tester mod changer function in that story.

valearna (approved):

PR #60, opened 2023-04-28
fixed code to allow curators to update/unset a license added by a cur…

sweng66:
…ator

### DIFF
--- a/src/components/biblio/BiblioFileManagement.js
+++ b/src/components/biblio/BiblioFileManagement.js
@@ -384,6 +384,7 @@ const FileEditor = () => {
                             onChange={(event) => patchReferencefile(referenceFile.referencefile_id, {"pdf_type": event.target.value === "" ? null : event.target.value}, accessToken)}>
                 <option></option>
                 <option>ocr</option>
+                <option>tif</option>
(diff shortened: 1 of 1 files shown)


src/components/biblio/BiblioFileManagement.js
```diff
@@ -338,7 +384,6 @@ const FileEditor = () => {
                             onChange={(event) => patchReferencefile(referenceFile.referencefile_id, {"pdf_type": event.target.value === "" ? null : event.target.value}, accessToken)}>
                 <option></option>
                 <option>ocr</option>
```
azurebrd, line 386: @sweng66 This tif option shouldn't be removed.  It probably happened from some merge oversight ?

azurebrd (changes requested):
Look great @sweng66 !  (just put back the "tif" <option> that shouldn't be removed, thanks !)

src/components/biblio/BiblioFileManagement.js
sweng66: Great catch!! Thanks @azurebrd 

Added back

azurebrd:
Great, thanks @sweng66 !

PR #61, opened 2023-05-01
added code to force the page to reload after adding or updating a lic…

sweng66:
…ense

### DIFF
--- a/src/components/biblio/BiblioFileManagement.js
+++ b/src/components/biblio/BiblioFileManagement.js
@@ -120,6 +120,7 @@ const OpenAccess = () => {
     }).then((res) => {
       setAlert("License Updated!");
       handleShowAlert(); // show the alert and hide it after 2 seconds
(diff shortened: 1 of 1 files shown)


azurebrd (commented):
I'll defer to @valearna here, I think he's done most of the FileManagement before ?  I'm not sure we want to reload, which might update what's in the redux store, or change the location in the page the user is viewing (I'm not sure it does either thing), as opposed to just doing an API call to get whatever could have been updated.

valearna:
@azurebrd I was thinking about the same thing, we should try to reload the data from the api instead of reloading the whole page, but this is ok if we need a quick and dirty solution for now

sweng66:
Makes sense! Thanks @azurebrd @valearna 

I will play around with api reloading and will ask for help if needed.

sweng66:
Will merge this change to main first and then fix it to reload the data from API.

PR #62, opened 2023-05-02
added code to force the section to reload after adding or updating a …

sweng66:
…license


azurebrd (approved):
Great, thanks for explaining that <Row key='open_access' {...(reloadSection ? { key: 'open_access_reload' } : {})}>  stuff @sweng66

PR #65, opened 2023-05-11
Scrum 2845

azurebrd:
Hi @valearna and @pjhale .  This code probably works and could be merged, but making a draft PR in case either of you is really opposed to how it's changed (I left a bunch of stuff commented out instead of removing it).  Basically, when someone logs on to okta, I'm mostly using Paul's code to get the MOD, then using the old code to get whether they're a developer, and additionally now checking for Tester and POTester, and storing those as values in the store, as siblings to oktaGroups.  Then every component that would previously import getOktaModAccess and use that, can just useSelector to get it from the store, and in most cases the access is based on oktaMod + oktaDeveloper, but in the Tracker it can just use oktaMod because we explicitly don't want developer there.  I think just storing some extra booleans and a string is going to be simpler than understanding and importing the function, which works differently in different cases, and since we have to change the background color based on whether the oktaMod is different from the manual tester Mod, it will probably be easier with the store (I haven't figured that out yet, but I don't think it'll be a problem).  Up next, I'll probably add the Tester component to save a tester Mod value to the store, and probably a 'different-than-oktaMod' boolean that would set a different background color.

Edit: Also fixed a key warning in Tracker.js and some import warnings on Biblio.js

### DIFF
--- a/.env.development_devserver
+++ b/.env.development_devserver
@@ -1,0 +1,13 @@
+# change the 3003 and 4003 to your ports (3006/4006, or whatever you use), don't change WDS_SOCKET_PORT from 443
+REACT_APP_OKTADOMAIN=dev-30456587.okta.com
+REACT_APP_CLIENTID=0oa1bn1wjdZiJWaJe5d7
+REACT_APP_UI_URL=https://dev3003.alliancegenome.org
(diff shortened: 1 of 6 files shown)


valearna:
Thanks @azurebrd, it looks good! I would remove commented out code to make the files more readable, but all changes make sense.

PR #67, opened 2023-05-17
scrum-1386 link main PDF to sort

FB-ping:
I am not sure if I got the access control here. please update whatever you think it is needed.

### DIFF
--- a/src/components/Sort.js
+++ b/src/components/Sort.js
@@ -9,7 +9,7 @@ import Col from 'react-bootstrap/Col';
 import Button from 'react-bootstrap/Button';
 import Alert from 'react-bootstrap/Alert'
 
-import { setReferenceCurie } from '../actions/biblioActions';
+import {downloadReferencefile, setReferenceCurie} from '../actions/biblioActions';
 import { setGetReferenceCurieFlag } from '../actions/biblioActions';
 
 import { changeFieldSortMods } from '../actions/sortActions';
@@ -22,7 +22,9 @@ import { closeSortUpdateAlert } from '../actions/sortActions';
 import { setSortUpdating } from '../actions/sortActions';
 import { sortButtonSetRadiosAll } from '../actions/sortActions';
 import {Spinner} from "react-bootstrap";
-
+import React, {useEffect, useState} from "react";
+import {searchXref} from "../actions/searchActions";
+import axios from "axios";
 
 // DONE
 // Find Papers to Sort will need to query data once there's an API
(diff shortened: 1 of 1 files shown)


FB-ping:
see https://dev3005.alliancegenome.org/sort for FB.   here I run with REACT_APP_RESTAPI=https://stage-literature-rest.alliancegenome.org, this expect to use the same database as https://stage-literature.alliancegenome.org/sort, however from my test UI I can see those two files I uploaded for release:  AGRKB:101000001001746  AGRKB:101000001001744, but can't see from stage sort UI.  Any idea ?

src/components/Sort.js
```diff
@@ -36,6 +38,42 @@ import {Spinner} from "react-bootstrap";
 
 const RowDivider = () => { return (<Row><Col>&nbsp;</Col></Row>); }
 
+export function getOktaModAccess(oktaGroups) {
```
azurebrd, line 41: Did you merge main into your branch ?  We shouldn't need getOktaModAccess anymore, since we can get it from the store, and we should also get it from the store and compare it to testerMod to use the testerMod value

azurebrd (changes requested):
We shouldn't be using oktaGroups and getOktaGroups anymore, with the testerMod code in main now.  We should also talk to Valerio about that referenceCurie passed to FileElement to see why it's an object.

src/components/Sort.js
FB-ping: @azurebrd @valearna which file have the most recent code as how to setup the access control so I can look at it ? BiblioFileManagement.js, BibliEditor.js, or Biblio.js  ?

FB-ping: @azurebrd this PR not merge into main yet, so I can't use those for now.

valearna (changes requested):
Okta access need to be resolved with the new changes that @azurebrd introduced and use of props without curly brackets need to be fixed

src/components/Sort.js
azurebrd: @FB-ping I'm pretty sure that PR is merged, it's got the Purple Merged label on the upper left.  Yeah, try what Valerio said, and merge main into your branch.  I linked the PR so you could see how I changed the code from the old way to the new way.

src/components/Sort.js
```diff
@@ -59,6 +83,72 @@ const Sort = () => {
...
+        const referencefileRow = (
+         <div>
+           {referencefileValue}
+        </div>);
+
+       rowReferencefileElements.push( referencefileRow ); }
+      return (
+        <>
+         {isLoading ?
+            <Row key="supplementExpandTogglerRow" className="Row-general">
```
azurebrd, line 145: This probably shouldn't be in a <Row><Col>  it could just be a span or maybe even just html text.  If it looks okay though, that's fine

azurebrd (approved):
That's great @FB-ping !  The code makes sense to me, and everything seems like it should work.  If it works, then cool !

src/components/Sort.js
azurebrd: @FB-ping my previous comment didn't show well because of markup.  Row and Col are probably not needed there, but if it looks okay, then cool.

PR #70, opened 2023-05-19
Order by Date SCRUM-2640

pjhale:
Moved actions to new file, added loading overlay, added to reducer.

Needs SCRUM-2640 PR in the API to work correctly.

### DIFF
--- a/src/actions/trackerActions.js
+++ b/src/actions/trackerActions.js
@@ -1,0 +1,56 @@
+import axios from "axios";
+
+const restUrl = process.env.REACT_APP_RESTAPI;
+
+export const searchMissingFiles = (mod_abbreviation, order_by) => {
+  return dispatch => {
+    dispatch(setLoadingState(true));
+    axios.get(restUrl + '/reference/missing_files/'+mod_abbreviation+"?order_by="+order_by)
+        .then(res => {
+          dispatch(setMissingFileResults(res.data));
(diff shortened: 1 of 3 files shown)


azurebrd (approved):
Great, thanks @pjhale . I'm glad you like actions and the reducer (Valerio sometimes prefers to do things in components, and I'm not sure when he's more okay with the store)

valearna:
My rule of thumb is to avoid adding things to redux if not strictly necessary to keep the store clean and more readable. In this case if you need to preserve the sorting parameters when navigating back and forth between pages then you must use redux but if you don't need to then you can use state variables within the component. The idea is to use useState as much as possible (see the first answer to this question: https://stackoverflow.com/questions/41584647/when-do-i-choose-react-state-vs-redux-store)

pjhale:
I would think in this case we would want to preserve this while users navigate pages.  I think the ideal workflow is something like ... user sees an entry on this page... user navigates to the biblio edit page or mod page to upload/search for data... user navigates back to this page to either refresh list or add a workflow tag.  Especially once we have pages I can see it being annoying if you were on page three and you are now looking for a specific paper that needs a workflow tag.

pjhale:
@valearna Bug fixed... should be good to merge whenever is convenient.

valearna (approved):

PR #74, opened 2023-05-31
Adding Filter

pjhale:
Default returns the normal data set... the other two return only  references that have a specific workflow tag.

### DIFF
--- a/src/components/Tracker.js
+++ b/src/components/Tracker.js
@@ -28,7 +28,7 @@ const DownloadButton = (mod) => {
       .then(blob => {
         const url = window.URL.createObjectURL(new Blob([blob]));
         const link = document.createElement('a');
-	let filename = `${mod.mod}_`;
+	let filename = `${mod.mod}_file_`;
 	if (trackerFilter == 'default') {
 	  filename = filename + "needed.tsv"
 	}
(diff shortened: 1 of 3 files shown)


pjhale:
Added in download button.  Moved the button out of the pagination element into its own element.  Made into a bootstrap button. @sweng66

pjhale:
This works... however the tag "_missing_files" is hardcoded into the UI here.  If you directly query the API you get the correct file name.

sweng66 (approved):
Looks great! Thanks @pjhale

src/components/Tracker.js
```diff
@@ -1,18 +1,64 @@
...
 import LoadingOverlay from "./LoadingOverlay";
 
+const DownloadButton = (mod) => {
+  const [downloading, setDownloading] = useState(false);
+  const orderBy = useSelector(state => state.tracker.orderBy);
+  const trackerFilter = useSelector(state => state.tracker.trackerFilter);
+
+  const handleDownload = () => {
+    setDownloading(true);
+    let apiUrl = `${process.env.REACT_APP_RESTAPI}/reference/download_tracker_table/${mod.mod}?order_by=${orderBy}&filter=${trackerFilter}`;
```
azurebrd, line 25: I thought mod was accessLevel which is a string, but if mod.mod works, then cool.

azurebrd: Cool, thanks @pjhale

PR #80, opened 2023-06-12
Adding checks to make sure certain SGD topics requires one or more db entities  

sweng66:


azurebrd (approved):
Nice @sweng66 !  I didn't know you were working on this.  I'm not sure what the curators want, but this code does what it says to do on the ticket https://agr-jira.atlassian.net/browse/SCRUM-2865 .  I would talk to curators about whether they want to validate that there's some text in the entity input, as opposed to at least one valid entity from the A-team lookup.  I would also ask whether they want the warning to go away automatically like it's doing or need manual dismissal, since stuff actively didn't go into the database.  Also whether they want a Modal instead of an extra Row, since most warnings I think have been in Modals (but we're definitely not consistent across a bunch of things in the UI components).  I think it's good though !

sweng66:
@azurebrd These are all great suggestions!! I will chat with Ceri about these tickets (SGD qualifier ones) after daily scrum so I will get some clarifications about these. I also have some questions for her. I will merge this into main first so I can show Ceri about this on stage. Thanks

PR #85, opened 2023-06-20
Scrum 2435

ianlongden:
species list not being reset on sort and lots of debug messages still need removing.


azurebrd (approved):
Good to know about the selected field in AsyncTypeahead and how to clear it.
Using the same list for mapping mod to taxon as the other component would probably be better than making two lists that work differently, but maybe we can eventually refactor common use things like that into a separate component.  And maybe we'll replace that with some db query for mod_to_taxon instead.
In general setting fontSize to a number em or a number rem is better than a number px, so that things scale when users zoom in/out, but it's up to curators how they like it.
Thanks @ianlongden and @valearna

PR #86, opened 2023-06-20
added code to automate qualifier addition for SGD when topics selected plus

sweng66:
* added code to get a list of qualifiers from A-team ATP ontology
* added code to make sure some topics (GO, phenotype, headline, protein complex, other primary literature) require entity
* also make sure Omics topics (HTP Phenotype data and other HTP data) do not require entity
* for review papers, the entity is optional, but they always get 'review display' 
* for other topics like 'pathway', 'regulation', PTM, gene model etc, the entity is optional, and they will get 'additional display' only if the entity is provided.

TODO: need to fix the warning message display. Hopefully, @valearna can lead a coding session to fix this along with other warnings across UI codebase.

### DIFF
--- a/src/actions/biblioActions.js
+++ b/src/actions/biblioActions.js
@@ -36,18 +36,6 @@ export const changeFieldArrayReferenceJson = (e) => {
   };
 };
 
-export const deleteFieldModReferenceReferenceJson = (e) => {
-  console.log('action delete field mod reference json ' + e.target.id + ' to delete');
-//   console.log(e);
-  const activeElement = getRevertButtonFromFontAwesomeElement(e.target);
-  return {
-    type: 'DELETE_FIELD_MOD_REFERENCE_REFERENCE_JSON',
-    payload: {
-      field: activeElement.id
-    }
-  };
-};
-
 export const changeFieldModReferenceReferenceJson = (e) => {
   console.log('action change field mod reference json ' + e.target.id + ' to ' + e.target.value);
 //   console.log(e);
@@ -56,18 +44,6 @@ export const changeFieldModReferenceReferenceJson = (e) => {
     payload: {
       field: e.target.id,
       value: e.target.value
-    }
-  };
-};
-
-export const deleteFieldCrossReferencesReferenceJson = (e) => {
-  console.log('action delete field cross references json ' + e.target.id + ' to delete');
-//   console.log(e);
-  const activeElement = getRevertButtonFromFontAwesomeElement(e.target);
-  return {
-    type: 'DELETE_FIELD_CROSS_REFERENCES_REFERENCE_JSON',
-    payload: {
-      field: activeElement.id
     }
   };
 };
@@ -85,23 +61,12 @@ export const changeFieldCrossReferencesReferenceJson = (e) => {
   };
 };
 
-export const deleteFieldModAssociationReferenceJson = (e) => {
-  console.log('action delete field mod association json ' + e.target.id + ' to delete');
-//   console.log(e);
-  const activeElement = getRevertButtonFromFontAwesomeElement(e.target);
-  return {
-    type: 'DELETE_FIELD_MOD_ASSOCIATION_REFERENCE_JSON',
-    payload: {
-      field: activeElement.id
-    }
-  };
-};
-
+// TODO to make live, add this to biblioActions.js  rename MOCK1_CHANGE_FIELD_MOD_ASSOCIATION_REFERENCE_JSON    create reducer action for it
 export const changeFieldModAssociationReferenceJson = (e) => {
   console.log('action change field mod association json ' + e.target.id + ' to ' + e.target.value + ' checked ' + e.target.checked);
 //   console.log(e);
   return {
-    type: 'CHANGE_FIELD_MOD_ASSOCIATION_REFERENCE_JSON',
+    type: 'CHANGE_FIELD_MOD_ASSOCIATION_REFERENCE_JSON',	// this doesn't do anything yet
     payload: {
       field: e.target.id,
       checked: e.target.checked,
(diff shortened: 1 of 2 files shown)


azurebrd (changes requested):
I think something's wrong, some actions are getting removed, but they probably shouldn't be. (e.g. deleteFieldModReferenceReferenceJson)  Maybe it happened from merging main into a local branch or something like that ?

src/actions/biblioActions.js
```diff
@@ -36,18 +36,6 @@ export const changeFieldArrayReferenceJson = (e) => {
   };
 };
 
-export const deleteFieldModReferenceReferenceJson = (e) => {
```
azurebrd, line 39: keep

src/actions/biblioActions.js
```diff
@@ -60,18 +48,6 @@ export const changeFieldModReferenceReferenceJson = (e) => {
   };
 };
 
-export const deleteFieldCrossReferencesReferenceJson = (e) => {
```
azurebrd, line 63: keep

src/actions/biblioActions.js
```diff
@@ -85,23 +61,12 @@ export const changeFieldCrossReferencesReferenceJson = (e) => {
   };
 };
 
-export const deleteFieldModAssociationReferenceJson = (e) => {
```
azurebrd, line 88: keep and don't add back the TODO comment

src/actions/biblioActions.js
```diff
@@ -85,23 +61,12 @@ export const changeFieldCrossReferencesReferenceJson = (e) => {
...
-    }
-  };
-};
-
+// TODO to make live, add this to biblioActions.js  rename MOCK1_CHANGE_FIELD_MOD_ASSOCIATION_REFERENCE_JSON    create reducer action for it
 export const changeFieldModAssociationReferenceJson = (e) => {
   console.log('action change field mod association json ' + e.target.id + ' to ' + e.target.value + ' checked ' + e.target.checked);
 //   console.log(e);
   return {
-    type: 'CHANGE_FIELD_MOD_ASSOCIATION_REFERENCE_JSON',
```
azurebrd, line 104: keep, don't add back the comment

src/components/biblio/BiblioEntity.js
```diff
@@ -257,29 +344,116 @@ const EntityCreate = () => {
...
+    if ( entityResultList && entityResultList.length > 0 ) {
+      for (let entityResult of entityResultList.values()) {
+        if (entityResult.curie === 'no Alliance curie') {
+          isEntityInvalid = true;
+	  break;
+        }
+      }
+    }
+    if (isPrimaryTopic) {
+      if (isEntityEmpty || isEntityInvalid) {
```
azurebrd, line 366: Do they want that error when any entity is invalid (meaning, everything they enter must be right), or when no entities are valid (meaning at least one entity must be right) ?  I can see the use-case for curators copy-pasting a bunch of stuff and not wanting to find all the things that are wrong and manually delete them.

azurebrd: Made a question to curators in the ticket

src/components/biblio/BiblioEntity.js
```diff
@@ -323,18 +497,108 @@ const EntityCreate = () => {
   const disabledEntityList = (taxonSelect === '' || taxonSelect === undefined) ? 'disabled' : '';
   const disabledAddButton = (taxonSelect === '' || taxonSelect === undefined) ? 'disabled' : '';
 
+  if (accessLevel === 'SGD') {
```
azurebrd, line 500: I still see the code for Entity and Topic Addition on old line 342  new line 606.  Does that mean that STD see something twice ?  (Just trying to understand)

azurebrd: I can't tell, is most of the code the same/similar ?  If so, maybe we could make components and reuse them, or have the if (accessLevel === 'SGD')  partially throughout displaying the row.  I'm not sure, I can't tell how similar or different it is.

src/actions/biblioActions.js
sweng66: Good catch!!! @azurebrd For some reason, I have a wrong version of biblioActions.js to start with. Just rebased code from main. So the missing code should be back in place now.

sweng66:
@azurebrd Thanks for looking over the code. Not sure why I started with a wrong version of biblioAction.js. Great catch! I have rebased code from main so this action file should be ok now.

sweng66:
@azurebrd about old line 342 vs new line 606, I just copied the old section for SGD papers,  just so I could fix the code to present a pulldown topic for SGD. It IS duplicate for now since I don't want to lose the old "search ahead code" for topic. 

I thought eventurally we may use the same section of code for all mods when MOD specific topic node is available. At that time, we can clean up the code to just have one set of code for all MODs or create MOD specific pages (each mod will have different taste and feel) and create reusable functions for each session (reusable small functions, eg, pulldown menu, type ahead search box, validation area, text box for note etc). 

Anyway, we can go over the code together as a group and decide how to organize everything including how to handle the warning messages.

Thanks Juancarlos!

sweng66:
@azurebrd Also currently, it requires all genes/entity names must be validated.

src/actions/biblioActions.js
azurebrd: Great, thanks !

azurebrd:
That's all great, thanks @sweng66 !

azurebrd (approved):
Sounds good about how it is now, and will defer to curators on validation, and we'll talk about code with Valerio when we all can :)

sweng66:
Sounds great! Thanks @azurebrd

PR #89, opened 2023-06-23
SCRUM-2872 new button in Merge section to discard changes, swap input…

azurebrd:
…, and re-query; which is distinct from just swapping the values in the form, which doesn't swap the right things when the winning paper doesn't have a PMID


azurebrd:
This branch name is wrong, but the commit has the right jira ticket

sweng66 (approved):
Looks good ! Thanks @azurebrd

azurebrd:
Thanks @sweng66 !

PR #90, opened 2023-06-26
adding ATP ID for 'other HTP data' and place holders for other topics that are not in ATP yet

sweng66:
…ther additional literature', 'other primary information'


azurebrd (approved):
Thanks @sweng66 .  So we'll just store those strings for now and sort out the data later ?  Or maybe not use it until the ATP values exist ?

sweng66:
@azurebrd Thanks for checking out the code.

Rob and Ceri agreed for now we will just store these as place holders (curators will not notice the difference. the UI will show everything in the normal way) until Ceri and Kimberly have time to sort these out (but we don't know when).

PR #91, opened 2023-06-27
Sort page has a button that shows only for accessLevel WB (test with …

azurebrd:
…Tester function), that pops up a modal that allows input of just the numeric part of an NCBITaxon ID to add to the ateam system through their GET endpoint at 'api/ncbitaxonterm/NCBITaxon:' + taxonId   If successful expect a response with res.data.entity.curie and display message.  Not sure what's really expected, for example tried 999 and that worked, but 9990 had a curie but no name, and asdf returned nothing, so using the last one as a failure and the middle as a success.


sweng66 (approved):
Looks good although I am not familiar with this sort page! Thanks @azurebrd

azurebrd:
Thanks @sweng66 !

PR #92, opened 2023-06-27
fix(crossReferences): use cross_reference_id instead of curie to acce…

valearna:
…ss cross ref to update


sweng66 (approved):
Looks good! Thanks @valearna

azurebrd (approved):
Great, thanks @valearna !

PR #93, opened 2023-06-28
fix(sort): using display: block to avoid overlap between UI elements

valearna:


valearna:
@azurebrd @pjhale there are multiple css issues that make the typeahead component disappear behind the row element. There is a fixed row height that prevents the row to expand when the component gets larger and all elements in the columns are `display: flex`, which is hard to manage. I created this draft PR with a possible solution. I removed all extra containers, set `display: block` for all columns and used line breaks to add more padding. Using css may be better but in this case I would avoid using `display: flex` to keep things simple

azurebrd:

azurebrd:
I just looked at the webpage with the changes, and I think it's fantastic.  Should me make it a full PR and merge into main @valearna ?

valearna:
Sounds good, thanks @azurebrd

PR #94, opened 2023-06-28
Make Alert element with position fixed to stay on the page and z-inde…

azurebrd:
…x high to display on top, and top 3.5 em to account for normal nav bar.  Possibly should ask curators about replacing the nav bar content with the not logged in bar instead and put the okta sign in there instead ?


valearna (approved):

azurebrd:
Thanks @sweng66 .  z-index is a css property that makes that element be on top of other elements, if their position would render them in the same place.  The higher number goes on top, so 9999 is an arbitrarily high number.

sweng66:
Oh cool. @azurebrd !!

PR #98, opened 2023-07-05
fix(sort): using new qualifier field + fixed loading overlay

valearna:


sweng66 (approved):
Looks great! Thanks @valearna

PR #100, opened 2023-07-10
fixed to always set 'Entity Type' pulldown to gene by default for SGD…

sweng66:
… even after hitting the 'Add' button

### DIFF
--- a/src/components/biblio/BiblioEntity.js
+++ b/src/components/biblio/BiblioEntity.js
@@ -234,10 +234,11 @@ const EntityCreate = () => {
   const [typeaheadOptions, setTypeaheadOptions] = useState([]);
   const [typeaheadName2CurieMap, setTypeaheadName2CurieMap] = useState({});
   const [warningMessage, setWarningMessage] = useState('');
-  
+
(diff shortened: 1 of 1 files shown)


valearna (approved):
I haven't tested the UI but the changes look good. I don't understand how the value is set in the store, but I guess it is set when you click the 'add' button? Thanks @sweng66.

sweng66:
Thanks both for checking out the code! 

@valearna Yes! 

@azurebrd The original fix worked great when first time curators use it.  But then it (Entity Type) got reset to null after hitting 'Add' button. I tried many ways to restore it to 'gene', but it wouldn't work. So I had to use this way (setting the state variable). In 95% cases, SGD curators are doing trigging for Gene so they prefer  to preset Entity type to 'Gene' (at least Rob likes this way). Also I have updated API to reset entity_type to None if an entity name is not provided (only for SGD) so it will work even though curators forgot to unset entity_type from 'gene' to null without a entity name. 

I also reset the topic to 'Pick a topic' after adding the entry into the database - basically reset everything to default just like first time adding things.

Maybe some time this week, we can look over the code together to see if we can fix this by calling dispatch action?

PR #102, opened 2023-07-10
fixed to do a dispatch action to update "entity_type" value instead of setting the state variable (thanks Juancarlos)

sweng66:


valearna (approved):

sweng66:
Thanks @valearna

PR #104, opened 2023-07-13
sgd code cleanup for topic entity tag UI

sweng66:


azurebrd (approved):
That's great @sweng66 !  I wonder if we could change the functions textArea and pulldownMenu to be entityTextArea and entityPulldownMenu, so there isn't ambiguity if we have other textareas or pulldownMenus later ?

azurebrd (commented):
Thanks @sweng66 .  I guess I didn't mean entity in the sense that it only applies to entities vs other things in this page.  I meant, textarea and pulldownMenu always dispatch changeFieldEntityAddGeneralField which only changes things in the biblioEntity component.

sweng66:
@azurebrd OK it totally makes sense!

sweng66:
Will change the names

sweng66:
fixed the two function names to 'tetTextArea' and 'tetPulldownMenu'. Thanks @azurebrd

azurebrd (approved):
Thanks @sweng66 !

PR #105, opened 2023-07-14
SCRUM-2492 A Team Warning

pjhale:
Warning for ateam api being down

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -437,6 +437,10 @@ em {
   visibility: hidden;
 }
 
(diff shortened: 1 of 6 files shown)


pjhale:
Should be good to go... had to resolve a merge conflict on the biblioedit page but it all looks fine.  Component now just needs to be imported and used in the return and everything else is taken care of.

valearna (approved):
Looks great, thanks @pjhale!

PR #107, opened 2023-08-02
entity addition of species list uses different a-team query because t…

azurebrd:
…heir data format is different


valearna (approved):

valearna:
thanks @azurebrd !

azurebrd:
Thanks @valearna !

PR #108, opened 2023-08-04
scrum-3135 add pagination at bottom

FB-ping:


azurebrd (commented):
I'll defer to @pjhale and @valearna that have worked on this more, but it seems like some of this is code duplication from SearchOptions.js and would be better exported from there and imported here, or into a general file for utils or something.  (I like keeping all the code in one file, so I'll defer to @valearna about how best to organize it)

pjhale:
I agree with @azurebrd that this code just needs to be re-used/exported and not duplicated.  I would also suggest that the bottom element should be inserted into the search layout page and not the search results page to be more consistent with the other elements and receive the correct formatting/padding.

FB-ping:
@pjhale I figure out how to set the layout as you suggested, see https://dev3005.alliancegenome.org/

pjhale (changes requested):
@FB-ping I think you missed adding the new SearchPagination file to git... this is not working for me right now.

FB-ping:
@pjhale sorry, you are right. Just add the file now.

pjhale:
Looks better.  There are a number of imports in this file that as far as I can see aren't used.  Form, setSearchSizeResultsCount, setSortByPublishedDate, Container, and Row should all be removed.

FB-ping:
> Looks better. There are a number of imports in this file that as far as I can see aren't used. Form, setSearchSizeResultsCount, setSortByPublishedDate, Container, and Row should all be removed.

@pjhale removed those not used imports, thanks

pjhale (approved):
Looks good now!

PR #110, opened 2023-08-14
Scrum 3045 - updated UI to use new topic entity tag APIs

valearna:
- get or create "curator" source at startup
- send data according to new schema when creating new topic entity tags


sweng66 (approved):
Looks good! Thanks @valearna

PR #113, opened 2023-08-23
ugly table to display API data in html table

azurebrd:
I realize this is an ugly html table (not responsive) and it's repeating some code, it's just a placeholder for Ceri to see what this kind of view would look like.

### DIFF
--- a/src/components/biblio/BiblioEntity.js
+++ b/src/components/biblio/BiblioEntity.js
@@ -39,6 +39,7 @@ export const curieToNameEntityType = { '': 'no value', 'ATP:0000005': 'gene', 'A
 const BiblioEntity = () => {
   return (<><AlertAteamApiDown />
           <EntityCreate key="entityCreate" />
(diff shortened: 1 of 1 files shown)


valearna:
Thanks @azurebrd, can we keep this on a separate branch and deploy it to one of the dev sites (e.g., dev3002)? I understand that this is important for Ceri to visualize all the possible columns but I'm afraid that if other curators see this they will automatically ask for the compact view without really thinking about how to display all the information in a meaningful way.

azurebrd:
@valearna I made it so it only displays that if your mod is ZFIN, so it'd only be an issue for someone that switches to ZFIN as tester, or for Holly.  I don't expect Ceri to look at this until next week, so if I left it deployed to 3002, I couldn't work on other stuff, since I develop there, but I don't really have other stuff to work on.  I could always switch to this branch when she's ready to test though.  Maybe talk it over ?

pjhale (approved):

PR #115, opened 2023-08-30
Scrum 3241 - new topic entity tag table (functional mockup)

valearna:
I think that this is now ready to be merged into main @azurebrd


azurebrd (approved):
Thanks @valearna !  
Passing the token to the API makes sense.  
I like the headersWithSortability set

PR #120, opened 2023-09-08
Ceri wants create to redirect to Biblio editor, not just to Biblio

azurebrd:


sweng66 (approved):
Looks good! Thanks @azurebrd 

BTW, can we import all functions at once from '../actions/biblioActions'?

azurebrd:
Thanks @sweng66 !  We could import them all at once, I just got used to have them in separate lines because it's easier to comment out and uncomment, instead of deleting and then having to look up what they were to put them back.  It's just a habit from when other people weren't working on the codebase.

sweng66:
@azurebrd Make sense! Thanks !!

PR #124, opened 2023-09-25
Scrum 2854

azurebrd:


valearna (approved):
Thanks @azurebrd this looks complicated! Maybe you can show me how it works on zoom?

PR #125, opened 2023-10-04
adding code to topic entity tag editor so it will allow users to filter by species

sweng66:


src/components/biblio/BiblioEntity.js
```diff
@@ -174,29 +223,116 @@ const EntityTable = () => {
...
-                                             setDescSort(true);
-                                           } else {
-                                             setSortBy(header);
-                                             setDescSort(!descSort)}
-                                         }}/> : null}</th>
-              )
-            } ) }
-            { source_headers.map( (header, index) => { return (<th key={`tetTableHeaderSource th ${index}`} >{header.startsWith('source_') ? header : 'source_' + header}</th>) } ) }
-          </tr>
+	  <thead>
```
valearna, line 226: Can we create a separate component for the filter? I'm sure this will be reused for other columns.

valearna (approved):
Looks good, thanks @sweng66. I left a comment asking to create a separate component for the filter but it doesn't need to be done now. We can always refactor the code when we need to add other filters.

src/components/biblio/BiblioEntity.js
sweng66: @valearna yes, we should. if not, the code will become so complicated if we are adding filtering for other columns. LOL

sweng66: again, it will be nice to work on it together in a pair/team programming session. @valearna

valearna: Sounds good. Let's merge these changes and then we can work in pair programming to refactor the code

PR #126, opened 2023-10-04
creating a new SpeciesFilter component and updating Topic/Entity/Tag editor to use SpeciesFilter

sweng66:


azurebrd (approved):
Thanks @sweng66 !

PR #127, opened 2023-10-10
topic/entity editor code clean up

sweng66:
1. pulled SGD topic/editor UI code out to have its own component
2. pulled PulldownMenu code to make it reusable 
3. pulled hard-coded species to TAXON mapping out to have its own file so multiple components can share the functions
    I am not sure we can get rid of this mapping since A-team contains a lot more species than we want to show in the tet UI.
    any way, we can get rid of this component when A-team has a list of species just for our UI.
4 cleaned up the tet related components to use these reusable modules


src/components/biblio/BiblioEntity.js
```diff
@@ -1,12 +1,28 @@
...
+        <TopicEntityTable key="entityTable" />
+      </>
+    );
+  };
+    
   return (
       <>
         <AlertAteamApiDown />
-        <TopicEntityCreate key="entityCreate" />
+        <TopicEntityCreateSGD key="entityCreate" />
```
azurebrd, line 25: This should still be the non-SGD TopicEntityCreate, right ?

azurebrd (commented):
Made a comment on one line in BiblioEntity.js  but in general will defer to @valearna since it's a refactor and I wouldn't have split things up.

src/components/biblio/BiblioEntity.js
sweng66: LOL good catch! @azurebrd 

Will fix it

valearna (approved):

PR #129, opened 2023-10-16
remove authorisation for restapi as ateam access now generated in api

ianlongden:
Need SCRUM-3314-api to be committed first.


azurebrd (approved):
Thanks @ianlongden

PR #131, opened 2023-10-18
SCRUM-3081 Adding Negation Facet

pjhale:
Introduces two styles of checkbox.  One with the standard check and another with both yes and no.


azurebrd (approved):
Thanks @pjhale !

PR #135, opened 2023-10-25
Sort should use same method to generate sources as TopicEntityTag Cre…

azurebrd:
…ate, says Valerio, using the same function, but the accessLevel for tet is different than for the rest of the sort page, to give developers access to referencefiles, so using a different variable.  Also, Kimberly says the entity_source should be alliance instead of manual.


sweng66 (approved):
Looks good! Thanks @azurebrd

azurebrd:
Thanks @sweng66 !

PR #136, opened 2023-10-27
Adding 'delete' button to the rows that login user has permission to delete in the Topic/Entity Editor

sweng66:


azurebrd (approved):
Thanks @sweng66 !

The console errors might be better as modal messages, and might want curator feedback on that.

Xenbase and XB sometimes need to be handled differently, but if you tested the XB case and it works, that's great.  (like, not sure if the lookup works, or what the accessLevel is, I imagine it's all XB for okta accessLevel + modCurie in db, but just in case one of them is 'Xenbase')

sweng66:
Good point about XB! will test

sweng66:
It works for XB too! Thanks @azurebrd

azurebrd:
cool @sweng66 !

PR #137, opened 2023-10-27
Add novel

ianlongden:
Needs to be added after api and alembic update. Works locally with this.


sweng66 (approved):
Looks good! Thanks @ianlongden

azurebrd (approved):

PR #141, opened 2023-10-31
new url for apteam check and align left on tick boxes

ianlongden:
use new url and look for specific data to pass. tick boxes aligned left on tet editor.


azurebrd (approved):
Looks good @ianlongden .  The curators might prefer the message to be the contents of something in there, like the data.json.message or data.status  but it's up to @cerivs to decide

PR #143, opened 2023-11-01
fixes to remove warning on compilation

ianlongden:
Fixes to remove warnings on compilation. Should be straight forward. 'warningMessage' is never set so no point having that code that checks for this and gives error message is the only thing that is not straight forward.

### DIFF
--- a/src/components/ATeamAlert.js
+++ b/src/components/ATeamAlert.js
@@ -39,7 +39,7 @@ const testAteamAPI = (accessToken, setStatus) => {
         .then(res => {
             const esindex = (element) => element.name === "Elasticsearch Indexing health check";
             const elasticsearch = res.data.checks.find(esindex);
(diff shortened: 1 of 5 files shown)


azurebrd (commented):
I'm very glad to have browser warnings removed, thanks @ianlongden .  I don't know what the warningMessage code being removed is.  I imagine at some point it had some functionality and someone removed it.  I'd rather check in with devs and curators about what it was about, and why it was removed before fully removing that code.

ianlongden:
Sure, no rush on this, more eyes the better. js all new to me so cleaning this up allowed me to review/see more code. Probably safe, but I am no expert, so yes lets see what others think.

valearna (approved):
Everything looks good to me. I would remove the unused code lines instead of commenting them out

PR #147, opened 2023-11-03
fix(topicEntityTag): Disable caching in AsyncTypeahead component

valearna:
The default cache option on AsyncTypeahead component was preventing duplicate API calls, and some logic was skipped. Now setting useCache property of the AsyncTypeahead component to false.


azurebrd (approved):
Thanks @valearna !

PR #149, opened 2023-11-06
fixed to reset the autocomplete/list switch option for entity_list

sweng66:
so only shows autocomplete option when 'species' is selected for topic


valearna (approved):
Looks good, thanks @sweng66 !

sweng66:
@valearna Thanks

PR #150, opened 2023-11-06
Use blue team api to get ateam status to avoid cors problems

ianlongden:


ianlongden:
Needs to be added after api, as it is dependent on it.

valearna (approved):

PR #153, opened 2023-11-08
Adjust axios POST request URLs in searchActions

valearna:
Fixed the URL format used for axios POST requests by appending a slash at the end within the 'src/actions/searchActions.js' file. This modification was done to align the URL structure with our backend server's routing conventions and avoid 307 errors from nginx


azurebrd (approved):
Thanks @valearna and @stuartmiyasato !

PR #154, opened 2023-11-08
Tidy up

ianlongden:
General removal of commented out code and old files. 
While commented out code does not hurt it makes it harder to find what you want and old is a valuable via git anyway.


src/components/ATeamAlert.js
```diff
@@ -16,7 +16,7 @@ export const AlertAteamApiDown = () => {
         if(accessToken) {
             testAteamAPI(accessToken, setStatus);
         }
-    }, [accessToken]); // eslint-disable-line react-hooks/exhaustive-deps
```
azurebrd, line 19: @ianlongden I forget what this was for, but I think I prevented new warnings showing up on the browser console.  It's like a # noqa  which is needed.

src/components/SwaggerUI.js
```diff
@@ -1,19 +1,15 @@
 import styles from '../index.css';
 
 import React, {useEffect} from 'react';
-// tslint:disable
```
azurebrd, line 4: This one and line 8 I don't know what they're for.   I don't think we use this component at all anymore.  It was imported in components/AppWithRouterAccess.js  and commented out  {/*<Route path = '/swaggerUI' component={SwaggerComp} />*/}

src/components/search/SearchOptions.js
```diff
@@ -26,25 +26,25 @@ const SearchOptions = () => {
                     </div>
                 </Col>
                 <Col sm={2}>
-                    <Form.Control as="select" id="selectSizeResultsCount" name="selectSizeResultsCount"
```
azurebrd, line 29: @ianlongden This looks like an actual change, not just cleaning up comments, so I don't know if that's on purpose from a different story, or an issue with rebasing.  When doing some recent API stuff, it was getting rid of some stuff that Shuai was working, so she rebased it.  So checking if this is your change on purpose or if it should talk to someone else about it.

azurebrd: Also the other changes on this file

src/reducers/mergeReducer.js
```diff
@@ -22,19 +22,6 @@ const initialState = {
     disableInput: '',
     blah: ''
   },
-//     input: 'AGR:AGR-Reference-0000569189',
```
azurebrd, line 25: These comments are helpful for telling people what to test when wanting to try different merge functionality, and are probably easier than keeping track of them in a separate document.

src/reducers/searchReducer.js
```diff
@@ -45,7 +45,7 @@ const initialState = {
   dateCreated: "",
   query_fields:"All",
   sort_by_published_date_order:"relevance",
-  partialMatch:"true"
```
azurebrd, line 48: @ianlongden I don't nkow how this works, sometimes a string of 'true' needs to be passed, but I don't know search.

azurebrd (commented):
@ianlongden There's a few comments on things that could use looking into.
Probably best if other people look at the things that are green in the diff.
Some of the green things are indenting changes, might be easier to put those back, unless the indenting is wrong.
I don't think the actions/index.js is needed, but sometimes an empty file is needed, as long as it runs, that's good.
The SwaggerUI component might be worth talking to Devs about it, and just removing it from the repo.

src/reducers/searchReducer.js
pjhale: Partial Match is what adds a wildcard after a users search.  The curators wanted this on by default (which is why its in the reducer).  This line still needs to be in the reducer.

src/reducers/mergeReducer.js
ianlongden: Okay if you find them useful I am happy to add them back.

src/components/search/SearchOptions.js
ianlongden: Sorry yes this is a change, we were getting warnings about 'selected' in pull downs and it said to declare this in the above with 'value'.  warnings now gone and we have the same functionality.

src/components/SwaggerUI.js
ianlongden: So this is probably okay?

src/components/ATeamAlert.js
ianlongden: There are no warnings being generated wrt this so guessing some thing else was removed previously that needed this.

azurebrd: Cool, sounds good

src/components/search/SearchOptions.js
azurebrd: Ah, so is 'value' just the starting value, and changing something into one of the options updating it ?  If the functionality works for changing from the dropdown, submitting, and getting the correct value instead of the original value, then cool

src/components/SwaggerUI.js
azurebrd: I guess :shrug:

src/reducers/mergeReducer.js
azurebrd: Cool, thanks :)

src/components/search/SearchOptions.js
ianlongden: We get the correct number of refs shown but the pull down does not reflect this , so something is awry.

ianlongden: Fixed, needed too use defaultValue not value.

azurebrd: Also for sortByPublishedDate ?

src/components/search/SearchOptions.js
```diff
@@ -26,25 +26,25 @@ const SearchOptions = () => {
...
                         <option value="10" >Results per page 10</option>
                         <option value="25" >Results per page 25</option>
-                        <option value="50" selected>Results per page 50</option>
+                        <option value="50" >Results per page 50</option>
                         <option value="100" >Results per page 100</option>
                     </Form.Control>
                 </Col>
                 <Col sm={2}>
-                    <Form.Control as="select" id="sortByPublishedDate" name="sortByPublishedDate"
+                    <Form.Control as="select" value="relevance" id="sortByPublishedDate" name="sortByPublishedDate"
```
azurebrd, line 42: this should be defaultValue @ianlongden

PR #159, opened 2023-11-27
Fixed the code to place 'Show More' / 'Show Less' at the right place for formatted abstracts

sweng66:


azurebrd (approved):
Thsnks @sweng66 !

PR #162, opened 2023-11-30
corpus sort source always disabled when mod not access level, whether…

azurebrd:
… new or from db


sweng66 (approved):
Thanks @azurebrd

azurebrd:
Thanks @sweng66 !

PR #165, opened 2023-12-07
Adding code to allow selection of number of rows displayed on table 

sweng66:


azurebrd (approved):
Thanks @sweng66 .  Does the page size stay if you go to another page in the App (like Biblio Editor) and come back ?   If so, that's great.  If not, I'd put the page number and page size in the store.  I guess it already doesn't keep the same page number when you change "tab", so this is consistent behaviour.  I just like keeping things in the store so that things are more consistent when switching pages.  I know @valearna doesn't like the store as much.

sweng66:
Will merge this into main first.

azurebrd:
THanks @sweng66 .  Might be good to ask curators, so you don't waste time if they explicitly don't want it, or other people think it'd be worse that way

PR #166, opened 2023-12-07
fixed to always use the URL submitted by MOD - through DQM submission…

sweng66:
… and it is stored in cross_reference.pages column


azurebrd (approved):
Nice @sweng66 !

sweng66:
Thanks @azurebrd

PR #167, opened 2023-12-07
Enable history navigation within biblio components

valearna:
Setting curie only through url parameter and not by the queryId component, which now pushes the ref curie to the history to load a new reference

### DIFF
--- a/src/actions/biblioActions.js
+++ b/src/actions/biblioActions.js
@@ -928,34 +928,6 @@ export const removeLoadingFileName = (fileName) => {
   }
 }
 
-export const queryId = (id) => {
-  return dispatch => {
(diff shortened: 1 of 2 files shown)


azurebrd (approved):
Thanks @valearna .  I'm not sure I follow.  Now when pressing the query button, it's doing a queryIdToCurie which sets the state variable for the curie, so the useEffect changes to the newUrl and pushes to the history.  Does pushing to the history make it a whole new page, and it reloads the values in the store, or is it just changing the page and the browser history, but keeping all the values ?  If it's keeping all the values, it's not changing state.biblio.referenceCurie is it, or is it changing it because location.search or location.pathname changed, so that triggers the useEffect that dispatches setReferenceCurie ?

I imagine it all works, so approving it, just asking because I don't think I follow the details

PR #168, opened 2023-12-11
fixed to preserve pagesize when navigating between tabs for the same paper and 

sweng66:
reset the pagesize to default value when switching to a different paper.

### DIFF
--- a/src/components/biblio/topic_entity_tag/TopicEntityTable.js
+++ b/src/components/biblio/topic_entity_tag/TopicEntityTable.js
@@ -25,7 +25,6 @@ const TopicEntityTable = () => {
   const biblioUpdatingEntityAdd = useSelector(state => state.biblio.biblioUpdatingEntityAdd);
   const referenceCurie = useSelector(state => state.biblio.referenceCurie);
   const storedPageSize = useSelector(state => state.biblio.pageSize);
-  const [previousReferenceCurie, setPreviousReferenceCurie] = useState(referenceCurie);
   const [totalTagsCount, setTotalTagsCount] = useState(undefined);
   const [page, setPage] = useState(1);
   const [sortBy, setSortBy] = useState(null);
@@ -101,12 +100,11 @@ const TopicEntityTable = () => {
     setShowSpeciesFilter(false);
   };
 
-  // const speciesInResultSet = new Set(topicEntityTags.map((tetDict) => tetDict.species));
   const speciesInResultSet = new Set(allSpecies);
-
+	    
   useEffect(() => {
     fetchDisplayTagData(accessToken);
-  }, [accessToken])
+  }, [accessToken]);
 
   useEffect(() => {
     const fetchMappings = async () => {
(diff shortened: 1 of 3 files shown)


src/components/biblio/topic_entity_tag/TopicEntityTable.js
```diff
@@ -32,9 +35,10 @@ const TopicEntityTable = () => {
   const [selectedSpecies, setSelectedSpecies] = useState([]);
   const [speciesFilterPosition, setSpeciesFilterPosition] = useState({ top: 0, left: 0 });
   const [allSpecies, setAllSpecies] = useState([]);
-  // set the initial pageSize state to 25
-  const [pageSize, setPageSize] = useState(25);
-    
+  const defaultPageSize = 25;
```
azurebrd, line 38: I don't think you need defaultPageSize in the component, because you're initializing it the reducer, and getting it from storedPageSize, so you could just use that instead of these 2 + pageSize

azurebrd (approved):
That's great, thanks @sweng66 !  I think there's an extra pair of variables because the redux store pageSize is initialized and you can just use that instead of 3 variables

azurebrd (commented):
@sweng66 Oh, actually, also the variables across all reducer files are shared, so instead of defaultPageSize calling it entityTablePageSize would avoid future conflicts better, unless curators want one page size across all pages, which Ceri mentioned, but I don't think they want that now.  Same with the cases SET_PAGE_NUMBER SET_PAGE_SIZE maybe better as SET_ENTITY_PAGE_NUMBER SET_ENTITY_PAGE_SIZE .  I do see with that, @valearna 's point that working with the store is more complicated

sweng66:
@azurebrd All your comments/suggestions make sense!! Will fix it accordingly. Thanks.

azurebrd:
Thanks @sweng66 !

src/components/biblio/topic_entity_tag/TopicEntityTable.js
sweng66: @azurebrd Fixed. =)

sweng66:
@azurebrd Just fixed the code to use variables specific to topic_entity_tag (TET) table. 
SET_TET_PAGE_SIZE etc..

src/components/biblio/topic_entity_tag/TopicEntityTable.js
```diff
@@ -35,10 +35,9 @@ const TopicEntityTable = () => {
   const [selectedSpecies, setSelectedSpecies] = useState([]);
   const [speciesFilterPosition, setSpeciesFilterPosition] = useState({ top: 0, left: 0 });
   const [allSpecies, setAllSpecies] = useState([]);
-  const defaultPageSize = 25;
-  //const [pageSize, setPageSize] = useState(defaultPageSize);
-  const pageSize = storedPageSize || defaultPageSize;
-  
+  //const defaultPageSize = 25;
+  //const pageSize = storedPageSize || defaultPageSize;
+  const pageSize = storedPageSize;
```
azurebrd, line 40: pageSize could just be directly set in line 27 from useSelector, but this is all great, thanks @sweng66 !

azurebrd (approved):
pageSize could just be directly set in line 27 from useSelector, but this is all great, thanks @sweng66 !

src/components/biblio/topic_entity_tag/TopicEntityTable.js
sweng66: you are right! lol @azurebrd

sweng66: @azurebrd Fixed. Sorry. back/forth so many times!

azurebrd (approved):
Thanks for your patience, @sweng66 !

PR #172, opened 2023-12-13
Changed validation column names

valearna:


valearna:
Thanks @sweng66 ! I'll merge this after restarting the API

PR #173, opened 2023-12-13
change display to show date_last_modified_in_pubmed and date_arrived_…

azurebrd:
…in_pubmed in a row, and date_last_modified_in_pubmed and date_created in another row

### DIFF
--- a/src/components/biblio/BiblioDisplay.js
+++ b/src/components/biblio/BiblioDisplay.js
@@ -206,6 +206,24 @@ export const RowDisplayCopyrightLicense = ({fieldIndex, fieldName, referenceJson
           <Col className={`Col-general ${cssDisplayRight} `} lg={{ span: 10 }}><a href={url} title={desc} rel="noreferrer noopener" target="_blank">{name}</a></Col>
         </Row>); }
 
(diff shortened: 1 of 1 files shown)


valearna (commented):
The two components look very similar, at least the jsx part. Can we create a single parametrized component instead of two?

azurebrd:
@valearna They're very similar, but in the BiblioDisplay section, I want to be able to switch without processing the data, and each of the components that get called there have a different way of processing.  I don't know what happened to the BiblioEditor changes, they should be part of this PR too, and they mostly mirror each other.  I think it's clearer to have the processing in each component instead of the parent component.  In the BiblioEditor, the updateFlag colours usually need processing too, but these two fields just happen to be display-only, so they don't need that logic, but in general we can't have a display of paired data, since those will need the updateFlag colour logic.  It seemed clearer to just make similar pairs of components

azurebrd:
@valearna I don't know how these BiblioEditor changes got into main, but they're here https://github.com/alliance-genome/agr_literature_ui/blob/main/src/components/biblio/BiblioEditor.js#L519  The Display part just mirrors that for consistency.

valearna:
Sounds good @azurebrd

valearna (approved):

azurebrd:
Thanks @valearna !

PR #174, opened 2023-12-14
Generalized RowDisplayPubmedPublicationStatusDateArrivedInPubmed and …

azurebrd:
…RowDisplayPubmedPublicationStatusDateArrivedInPubmed to work for BiblioEditor and BiblioDisplay, since they're displayOnly fields.  Generalized displaying a pair of data in one row, as Valerio suggested.  Added display of RowDisplayBooleanDisplayOnly for prepublication_pipeline, generalized to allow any future boolean field that should display its tag as the value, when the value it true


valearna (approved):
looks great, thanks @azurebrd

azurebrd:
Thanks @valearna !

PR #177, opened 2023-12-15
Adding Mod Preferences

pjhale:
When someone is logged in the facet "in corpus or needs review" is automatically selected.  This is accomplished by another variable in the reducer that tracks is we have loaded these yet.  Users can deselect the preloaded facet and it will only return on a page refresh.

Keeping as draft until Kimberly can look at it.


azurebrd (approved):
Thanks @pjhale !

PR #178, opened 2023-12-15
adding AGMs/subtypes to entity_type pulldown and adding code to validate these new entity types/subtypes

sweng66:


valearna (commented):
It looks like there's a lot of hardcoded mappings. Is there a way to do this without hardcoding @sweng66 ?

azurebrd (approved):
Looks fine to me, but I think you're both meeting to discuss how to do this with less hardcoding

PR #180, opened 2023-12-19
modal for notes that are over 100 characters, table header has 16em m…

azurebrd:
…in width

### DIFF
--- a/src/components/biblio/topic_entity_tag/TopicEntityTable.js
+++ b/src/components/biblio/topic_entity_tag/TopicEntityTable.js
@@ -42,7 +42,10 @@ const TopicEntityTable = () => {
   };
   const [selectedCurie, setSelectedCurie] = useState(null);
   const [showModal, setShowModal] = useState(false);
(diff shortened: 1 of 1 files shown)


azurebrd:
@sweng66 The NoteModal is like your CuriePopup.  Would you be okay if I merge them both into one component GenericTetTableModal that also takes in the Modal.Title value as a parameter, then we can avoid having similar components, or do you think it's better to have it explicit for each type of Modal ?

sweng66:
@azurebrd Sounds great to have GenericTetTableModal so we don't need to repeat code!

sweng66 (approved):
Looks good! Thanks @azurebrd

azurebrd:
Great, thanks @sweng66 , I've made the change

sweng66:
@azurebrd Saw it. Great.

sweng66 (approved):

azurebrd:
Great, thanks @sweng66 !

PR #181, opened 2023-12-19
Added 'null' value to 'negated' field and changed checkbox into select

valearna:
I want to show this to Kimberly at standup before merging


azurebrd (approved):
Thanks @valearna

PR #184, opened 2023-12-21
When creating new topic entity data from an entity list, if the same …

azurebrd:
…term is input by curators multiple times, subsequent times return 'duplicate' in grey instead of the curie it maps to

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -229,6 +229,10 @@
   background-color: var(--light-red);
 }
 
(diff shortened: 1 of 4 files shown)


valearna (approved):
I tested it locally and checked the code and everything looks great, thanks @azurebrd!

azurebrd:
Thanks @valearna !

PR #185, opened 2023-12-22
reverse priority in case of someone not being logged on having access…

azurebrd:
…Level be No, but the file being open access true, which was creating links for download, which could not be downloaded


valearna (approved):
Thanks @azurebrd!

azurebrd:
Thanks @valearna !

PR #192, opened 2024-01-12
SCRUM-3386 Removing Whitespace from searches

pjhale:
Search now trims whitespace from both ends before sending to ES.  This prevents weird behavior that happens with wildcards.  Also changed so that ES always gets the search string after the translation of AND's and OR's.


valearna (approved):
Thanks @pjhale!

PR #193, opened 2024-01-17
Adding code to topic/entity UI to check and display the existing tags (for case 1 + 5) 

sweng66:


azurebrd (approved):
Thanks @sweng66 !  (Neat, hadn't seen Promise resolve reject like that before)

PR #198, opened 2024-01-29
fixed the total tag count bug when deleting a tag

sweng66:
… tag count will be refreshed automatically


valearna (approved):
Thanks @sweng66 !

sweng66:
Thanks @valearna

PR #200, opened 2024-01-29
fixed to only show the duplicate tag alert section if there is any du…

sweng66:
…plicate tags available


azurebrd (approved):
Thanks @sweng66

sweng66:
@azurebrd Thanks!

PR #203, opened 2024-02-05
adding code to send an alert when an invalid curie or XREF ID is entered for biblio se…

sweng66:
…arch/editor pages

### DIFF
--- a/src/components/Biblio.js
+++ b/src/components/Biblio.js
@@ -411,7 +411,8 @@ const BiblioIdQuery = () => {
   const history = useHistory();
   const [idQuery, setIdQuery] = useState('');
   const [error, setError] = useState(''); // state to store the error message
-  
+  const [showAlert, setShowAlert] = useState(false);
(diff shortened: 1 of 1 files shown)


azurebrd (changes requested):
Since you're importing the Alert component, could you use it instead of a <div> ?

Like 
ATeamAlert.js:            <Alert className="ateam-alert" variant="danger" onClose={() => setStatus(true)} dismissible>
ATeamAlert.js:                <Alert.Heading>Looks like {ateamApiBaseUrl} is down</Alert.Heading>
ATeamAlert.js:            </Alert>

Which lets you set "dismissable" so curators can close the alert by clicking the x

sweng66:
Cool @azurebrd Will fix

azurebrd (approved):
Great, thanks @sweng66 !

sweng66:
Cool. Thanks @azurebrd

PR #205, opened 2024-02-09
construct code

ianlongden:


src/actions/biblioActions.js
```diff
@@ -13,6 +13,16 @@ const restUrl = process.env.REACT_APP_RESTAPI;
 //const ateamApiBaseUrl = 'https://beta-curation.alliancegenome.org/';
 const ateamApiBaseUrl = process.env.REACT_APP_ATEAM_API_BASE_URL;
 
+const taxon_to_mod = {"NCBITaxon:559292": "SGD",
```
azurebrd, line 16: There's a components/biblio/topic_entity_tag/TaxonUtils.js  This might be better there and imported ?  I'd welcome @valearna and @sweng66 feedback on this.   Possibly a function that works on getModToTaxon to reverse the mappings and generates this would be better.

azurebrd (changes requested):
@valearna @sweng66 There's a comment that I'd like your feedback on.

@ianlongden a lot of the changes are reformatting, but I saw a few sections that seem new, and they seem fine.  The construct stuff is in a different format, so it needs new structure based on uniqueId ?

src/actions/biblioActions.js
sweng66: @ianlongden Agree with @azurebrd. We should get this mapping from the database by calling endpoint `${process.env.REACT_APP_RESTAPI}/mod/taxons/all` or calling functon "getModToTaxon" in src/components/biblio/topic_entity_tag/TaxonUtils.js / reverse the mapping / adding Homo sapiens to the mapping after.

sweng66: @ianlongden Agree with @azurebrd. We should get this mapping from the database by calling endpoint ${process.env.REACT_APP_RESTAPI}/mod/taxons/all or calling functon "getModToTaxon" in src/components/biblio/topic_entity_tag/TaxonUtils.js / reverse the mapping / adding Homo sapiens to the mapping after.

sweng66:
@ianlongden can we pass an optional parameter (taxonToMod) to function "changeFieldEntityEntityList" from TopicEntityCreate component?

taxonToMod = modToTaxon.fromEntries(modToTaxon.entries(obj).map(([key, value]) => [value, key]));

The modToTaxon is already set in TopicEntityCreate.

src/actions/biblioActions.js
valearna: using `getModToTaxon` sounds good.

src/components/biblio/topic_entity_tag/TopicEntityCreate.js
```diff
@@ -66,16 +67,20 @@ const TopicEntityCreate = () => {
     
   const [curieToNameTaxon, setCurieToNameTaxon] = useState({});
   const [modToTaxon, setModToTaxon] = useState({});
+  const [taxonToMod, setTaxonToMod] = useState({});
```
azurebrd, line 70: @ianlongden : I think @sweng66 's function to remap modToTaxon -> taxonToMod would go here.  Modified to loop through every mod's taxon instead of just the first taxon

azurebrd (changes requested):
See inline comment about Shuai's function

sweng66:
previous one is incorrect. the following code should work:

const taxonToMod = {};
for (const [mod, taxons] of Object.entries(modToTaxon)) {
  taxonToMod[taxons[0]] = mod;
}

azurebrd:
Well, in case they ever do have functions, maybe (I haven't tried it, but I think it makes sense)
const taxonToMod = {};
for (const [mod, taxons] of Object.entries(modToTaxon)) {
  taxons.forEach((taxon) => { 
     taxonToMod[taxon] = mod;
  });
}

sweng66:
@azurebrd yes, this should cover everything !!

PR #206, opened 2024-02-12
Fixed the search UI to make the facets section always visible 

sweng66:
Fixed the search UI to make the facets section always visible on the left side of the screen as users scroll down the search results page. Also added code to make facets section scrollable.

Should we put the css code for this in App.css?


azurebrd (approved):
Nice @sweng66 !  Up to you about putting the css in App.css   At first I tried to put a lot there thinking we'd be using that for various things, but as different people worked on different pages, they went with different styling, so I don't know.

sweng66:
@azurebrd Got it. I will just leave the code as is. Thanks!

PR #212, opened 2024-02-13
Scrum 3453 2

ianlongden:
I think I got it this time.

### DIFF
--- a/src/actions/biblioActions.js
+++ b/src/actions/biblioActions.js
@@ -370,7 +370,7 @@ export const updateButtonBiblioEntityAdd = (updateArrayData, accessLevel) => {
 
 
 
-export const changeFieldEntityEntityList = (entityText, accessToken, taxon, entityType, taxonToMod) => {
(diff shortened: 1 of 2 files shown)


sweng66 (commented):
Thanks @ianlongden.

It looks great! Can you fix the "changeFieldEntityEntityList"  function to make the taxonToMod as an optional parameter? something like:

export const changeFieldEntityEntityList = (entityText, accessToken, taxon, entityType, taxonToMod = undefined) => {

}

This function is used by other component (eg TopicEntityCreateSGD,js) and this component doesn't need to pass taxonToMod.

Also using @azurebrd code to get the mapping?

const taxonToMod = {};
for (const [mod, taxons] of Object.entries(modToTaxon)) {
  taxons.forEach((taxon) => { 
     taxonToMod[taxon] = mod;
  });
}

sweng66 (approved):
Looks good. Thanks @ianlongden

PR #216, opened 2024-02-29
merge data transfer accounts for relations between two merging refere…

azurebrd:
…nces, in which case both are red for removing, and while transferring data it's removed from the losing reference list, it's the same connection so does not need to be removed from the winning reference side.  refactor toggle to use boolean.


sweng66 (approved):
Looks good! Thanks @azurebrd

azurebrd:
Thanks @sweng66 !

PR #217, opened 2024-03-01
store modEntityId for constructs and not name

ianlongden:
Blooming slow due to API call being super slow!! But it works. Needs agr_literature_service updated first to see transform.


sweng66 (approved):
Looks good! Thanks @ianlongden

PR #218, opened 2024-03-01
Integrating AG Grid into the table

pjhale:
WIP

### DIFF
--- a/src/components/biblio/topic_entity_tag/TopicEntityTable.js
+++ b/src/components/biblio/topic_entity_tag/TopicEntityTable.js
@@ -1,5 +1,5 @@
 import {useSelector, useDispatch} from "react-redux";
-import {useEffect, useState} from "react";
+import {useEffect, useState, useMemo, useCallback, useRef} from "react";
 import {fetchDisplayTagData} from "../../../actions/biblioActions";
 import { setTetPageSize as setPageSizeAction } from "../../../actions/biblioActions";
 import axios from "axios";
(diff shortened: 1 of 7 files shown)


src/reducers/biblioReducer.js
```diff
@@ -980,7 +980,7 @@ export default function(state = initialState, action) {
         referenceCurie: action.payload,
         entityAdd: setRefCurieDefaultEntityAdd,
 	    tetPageSize: defaultTetPageSize,
-        curieToNameTaxon: [],
+        curieToNameTaxon: {'NCBITaxon:7955': 'Danio rerio', 'NCBITaxon:8364': 'Xenopus tropicalis'},
```
azurebrd, line 983: Why do we need these values ?  I thought they should come from components/biblio/topic_entity_tag/TaxonUtils.js getCurieToNameTaxon

pjhale: I need to remove this... this was for testing.

src/components/biblio/topic_entity_tag/TopicEntityTable.js
```diff
@@ -23,9 +15,9 @@ import "ag-grid-community/styles/ag-theme-quartz.css"; // Theme
 const TopicEntityTable = () => {
   const dispatch = useDispatch();
   const accessToken = useSelector(state => state.isLogged.accessToken);
-  const oktaMod = useSelector(state => state.isLogged.oktaMod);
-  const testerMod = useSelector(state => state.isLogged.testerMod);
-  const accessLevel = (testerMod !== 'No') ? testerMod : oktaMod;  
+  //const oktaMod = useSelector(state => state.isLogged.oktaMod);
+  //const testerMod = useSelector(state => state.isLogged.testerMod);
+  //const accessLevel = (testerMod !== 'No') ? testerMod : oktaMod;
```
azurebrd: I guess we don't need okta/access because the delete functionality is being removed ?  I've lost track of what it should do, so if curators no longer want that, ok.

pjhale: The delete function is now in another file... so we dont need it here.  All that code is in TopicEntityTagActions.jsx.  That is the cell renderer that builds the values in that cell.

azurebrd: cool, thanks

src/reducers/biblioReducer.js
azurebrd: cool cool

pjhale: This has been removed...

valearna (approved):
This looks awesome, thanks @pjhale !

PR #220, opened 2024-03-07
Scrum 3774 - new TET source model

valearna:
- Updated function to retrieve default source for ABC + mod and changed default data to reflect new TET source model
- Updated TET table with new fields


sweng66 (approved):
Looks good! Thanks @valearna

PR #221, opened 2024-03-08
Scrum 3402 fixes

pjhale:
Removes more code and uses names from the API return and not direct from A team.  Also fixes some bugs from adding and removing code.

### DIFF
--- a/src/components/AgGrid/TopicEntityTagActions.jsx
+++ b/src/components/AgGrid/TopicEntityTagActions.jsx
@@ -9,7 +9,7 @@ export default (props) => {
     const accessLevel = (testerMod !== 'No') ? testerMod : oktaMod;
 
     const handleDeleteClick = async () => {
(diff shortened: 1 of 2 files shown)


src/components/AgGrid/TopicEntityTagActions.jsx
```diff
@@ -39,7 +39,7 @@ export default (props) => {
 
     return (
     <span>
-        {props.data.topic_entity_tag_source.mod === accessLevel ? <button onClick={() => handleDeleteClick()}>Delete</button> : null}
+        {props.data.topic_entity_tag_source.data_provider === accessLevel ? <button onClick={() => handleDeleteClick()}>Delete</button> : null}
```
valearna: Same here, we should use `secondary_data_provider_abbreviation`

pjhale:
@valearna Changed the mod value for the delete button code.

valearna (approved):
Thanks @pjhale !

PR #223, opened 2024-03-14
Fixing Author Filter and Code Cleanup

pjhale:
Author filter was broken because the function did not have the exclude facet value.  This is because we pass in a huge list of variables for every ES call... but now we grab all those from state instead.  Also created a function to build search params values so that this code is not duplicated.

This also fixes display tag name since we did that in the AG Grid demo.


valearna (approved):
Looks good, thanks @pjhale !

PR #231, opened 2024-03-29
Scrum 3673 topic entity delete warning

FB-ping:


src/components/AgGrid/TopicEntityTagActions.jsx
```diff
@@ -1,20 +1,44 @@
...
             console.error("Permission denied. Cannot delete this row.");
             return;
         }
+        let topic=props.data.topic_name;
+        let entity_type=props.data.entity_type_name;
+        let entity=props.data.entity_name;
+
+        let noData=props.data.negated;
+        let novelData=props.data.novel_topic_data;
+        setBody("topic:" + topic+" entity type:"+entity_type + " entity:"+ entity + " no Data:"+noData+ " Novel Data:" + novelData);
```
pjhale, line 35: I would like to see this data on different lines so that its easier to read rather than all on one line

src/components/AgGrid/TopicEntityTagActions.jsx
```diff
@@ -28,18 +52,38 @@ export default (props) => {
...
             console.error("Error deleting item:", error);
         }
+        setShowModal(false);
+     }
     };
 
 
     return (
     <span>
-        {props.data.topic_entity_tag_source.secondary_data_provider_abbreviation === accessLevel ? <button onClick={() => handleDeleteClick()}>Delete</button> : null}
```
pjhale, line 42: The way that this used to work with a ? operator was a bit more explicit on what exactly is being shown.  Instead now the button is 10ish lines of code down.  Not sure if there is a reason for this.

src/components/AgGrid/TopicEntityTagActions.jsx
FB-ping: @pjhale good suggestion, fixed now.

src/components/AgGrid/TopicEntityTagActions.jsx
FB-ping: @pjhale I changed back to original way.

PR #232, opened 2024-04-01
adding code to take care of special characters

sweng66:


valearna (approved):
I don't know much about these changes but they look good to me. Thanks @sweng66

PR #237, opened 2024-04-03
fix bug for format true/false value

FB-ping:


azurebrd (commented):
@FB-ping does that work ?  It seems to be asking that if null, then this or that, but null is always null, so never true.  Do you want instead ?
let noData = (noData === null) ? this : that

FB-ping:
@azurebrd you are right, I just want to check if negated is null or not (  it could be null, false, or true).  updated the code and test some case with negated ==null. so it should work fine now.

PR #238, opened 2024-04-03
fixed to use case-insensitive sorting

sweng66:


azurebrd (approved):
Oh, that's neat, I wouldn't have thought of the null cases, cool.  Thanks @sweng66

sweng66:
@azurebrd Thanks!

PR #241, opened 2024-04-08
Scrum-3603 Validation by curator

pjhale:
Adds the validation by curator buttons.  

This PR does not use the UID yet but has it stored in the reducer.

There are issues with tags actually getting validated.


pjhale:
Issues resolved.  Now using new force_insertion param.  Fixed empty modal bug. Better formatting.  Ready for review/merge.

valearna (approved):
Everything looks good to me, thanks @pjhale!

PR #242, opened 2024-04-08
fixed to handle new parameter force_insertion

sweng66:


valearna:
Thanks @sweng66 , I'll defer to @pjhale since he's working on changes that use the 'force insertion' option

sweng66:
Okay thanks @valearna

pjhale:
This should be fine... just to be clear I should still pass the curie with an additional JSON parameter "force_insertion: true".  Correct?  With the curator function I think we always want to force the insertion.

sweng66:
Currently, API will check if the tag row has force_insertion passed in (along with the reference curie), if yes and it is true, then it will not check for duplicate tag row(s), instead it will insert the new row into the database directly. 

I don't really know what you are using the force_insertion parameter for so I have no idea if my fix will affect your work or not. @pjhale 

We can talk about it at standup.

valearna (approved):

sweng66:
Thanks @valearna

PR #243, opened 2024-04-09
adding code to skip obsolete ATP topic terms; also mark the obsolete …

sweng66:
…entity during entity validation


azurebrd (approved):
Thanks @sweng66 !

PR #245, opened 2024-04-10
Added checks to make sure specific cookie exists before manipulating it

valearna:


valearna:
Thanks a lot @azurebrd and @sweng66 for the debugging session!

PR #246, opened 2024-04-10
adding code to make construct work for both name and symbol as well as ID

sweng66:


azurebrd (approved):
Thanks @sweng66 !

sweng66:
Thanks @azurebrd

PR #248, opened 2024-04-13
getXrefPatterns from actions into store when BiblioEditor loads, from…

azurebrd:
… API endpoint to get all reference xref patterns.  validateXref pattern when changing prefix dropdown, or blurring curie input.


sweng66 (approved):
Looks good! Thanks @azurebrd 

Will the validateXref function be reused somewhere else? maybe not. but if yes, then it might be better placed outside the component.

azurebrd:
:nod:  thanks @sweng66 .  Eventually we might have a resource editor, and need to use it there, but I think this is the only place that allows editing or inputting xref curies.  The current xref resource patterns are just copies of the reference ones, not even real ones that curators have come up with.

sweng66:
makes sense! @azurebrd

PR #250, opened 2024-04-15
Scrum 3586 custom topic entity table

FB-ping:
changes from src/actions/biblioActions.js and src/components/biblio/topic_entity_tag/TaxonUtils.js are for scrum-3836 'search filter for A team",  follow @valearna 's instruction.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,10 +36,11 @@
     "redux-thunk": "^2.3.0",
     "web-vitals": "^1.1.2",
     "ag-grid-community": "~31.0.2",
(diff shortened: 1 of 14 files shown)


valearna:
I don't know what the package mobx does but the code looks good to me. I'll defer to @pjhale

azurebrd (approved):
Looks great to me, thanks @FB-ping !

If @pjhale or @valearna think we don't need the container and rows to handle the left float, we can do it however they think instead, too.  I still haven't done anything with AgGrid

valearna (approved):

PR #254, opened 2024-04-20
use cookie to store setting of hide/show columns

FB-ping:


azurebrd (changes requested):
I thought we were already using cookies.  That @pjhale had code for the Topic Entity Tag stuff.  We should probably use the same cookie, or type of cookie storage, instead of a new module.  But we can see what the Paul, @valearna and @sweng66 think

sweng66:
I am not familiar with cookies, but to manage the column visibility states without using the library, we probably can use the browser's native localStorage (or sessionStorage). These storage APIs provide a way to store data across browser sessions (in the case of localStorage). Just googled the "react-cookie". It seems like there are some good parts about it. It simplifies the process of managing cookies, making it easier to set, retrieve, and react to changes in cookie values. I am not sure if it is overkill to use it in our case (e.g., just storing a simple preference like column order).

azurebrd:
@FB-ping I haven't used this code, but https://github.com/alliance-genome/agr_literature_ui/blob/main/src/components/biblio/topic_entity_tag/TopicEntityTable.js#L323 has the way it's handling the cookie now.  I don't know whether line 323 sets the whole cookie to have that data, or whether it updates that field within the cookie.  If you change your code to use the same cookie, I think it's javascript object, and you can use a new field name for your data.  Just check that 323 doesn't wipe your data, and that whatever you add doesn't wipe columnOrder from the cookie.  You can see how the cookie data is accessed in 333

FB-ping (commented):
use same cookie as Paul's now.

pjhale (approved):
I'm still having some issues testing but as far as I can tell this works fine.  When i tested on 3005 it was working as I would expect.

PR #256, opened 2024-04-25
update code to work with nested object type for topic_entity_tags and tet single tag search option

sweng66:
* adding "apply selections to single tag" checkbox
* adding a warning message when multiple topics or multiple confidence levels are selected and the "apply selections to single tag" checkbox is checked
* adding code in searchActions.js to preprocess facets selection data to construct nested search data structure


sweng66:
Thanks @valearna

PR #257, opened 2024-04-26
Only allow uploading files if accessLevel is not No, for cases like C…

azurebrd:
…hris Grove that purposely are not in a curator group on Okta


sweng66 (approved):
Looks great! Thanks @azurebrd

azurebrd:
Thanks @sweng66 !

PR #258, opened 2024-04-30
someone with okta credentials not belonging to a mod could try to acc…

azurebrd:
…ess this and would break when accessing modFileStatus[accessLevel], so check and return if it would break


sweng66 (approved):
Looks good! Thanks @azurebrd

azurebrd:
Thanks @sweng66 !

PR #260, opened 2024-05-04
add restore default setting choice

FB-ping:


sweng66 (approved):
Your "handleSelectDefault" function makes sense @FB-ping. 

Is it possible to get rid of the redundant code (eg, the conditions i && i.checked == true and i && i.checked == false are kind of redundant? also newItems? ).  Something like the following:

const handleSelectDefault = () => {
    setItems(itemsInitOrg);
    const columnStates = itemsInitOrg.map(item => ({
        colId: item.field,
        hide: !item.checked
    }));

    gridRef.current.api.applyColumnState({
        state: columnStates,
    });

    const newItemsStr = JSON.stringify(itemsInitOrg);
    document.cookie = `items=${newItemsStr}; expires=Thu, 18 Dec 2050 12:00:00 UTC; SameSite=None; Secure`;
};

I haven't tested out the code. Just an idea!

FB-ping:
@sweng66 good catch, actually I am also very frustrated here, I originally with exact code as you mentioned here, and found out that hide: !item.checked didn't work at all (no idea why not work as we expected), and it only work with true or false here. That's why I have to use the if statement here.

azurebrd:
That's weird.  Both itemsInit and itemsInitSGD all have values that have a checked set to a boolean.  It's looping through each 
`newItems.forEach(i => {`
so I wouldn't think you'd need to check that i exists, because newItems only has the hardcoded values
`if (i && i.checked == true)`
And you have another variable, that presumably works in getting a reverse of the boolean ?
`let columChecked = !i.checked;`
(Also, that variable isn't used elsewhere in the PR, so maybe it's not needed ?  Also, it's missing the final n on column)
I dunno.

PR #261, opened 2024-05-07
refactored nested query code

sweng66:


valearna (approved):
Topic and confidence_level are hardcoded here. It's ok for now but we should try to generalize the code for all other TET fields in the future

sweng66:
Sounds good! Thanks @valearna

PR #262, opened 2024-05-08
Scrum 3932 topic entity tag width set

FB-ping:


sweng66 (approved):
Looks good to me! Thanks @FB-ping

PR #267, opened 2024-05-12
completeMergeReferences should not hide MergeDataTransferModal, inste…

azurebrd:
…ad when MERGE_BUTTON_API_DISPATCH reducer returns, it should set that to hide.  Also MergeDataTransferredModal can show when dataTransferHappened and mergeTransferringCount is zero, or also when mergeCompletingCount is not zero, because the merging is still happening, but the modal hasn't come up yet.  This is confusing, but it works and fits the story

### DIFF
--- a/src/components/Merge.js
+++ b/src/components/Merge.js
@@ -172,7 +172,7 @@ const MergeSelectionSection = () => {
       ) }
     })()}
 
-    { dataTransferHappened && mergeTransferringCount === 0 ?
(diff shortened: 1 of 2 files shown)


sweng66 (approved):
Looks good! Thanks @azurebrd

sweng66:
It sounds complicated to me.  =)

azurebrd:
haha, thanks @sweng66 !

PR #268, opened 2024-05-14
Style fixes for SCRUM-3299

pjhale:
Making sure the buttons always render on their own line.  Also adding padding to the bottom of the facet to give a bit more room between date elements.


azurebrd (approved):
Thanks @pjhale !

PR #270, opened 2024-05-17
Adding Entity Filter

pjhale:
Adds the entity filter.  Functions almost exactly the same as the species filter but we dont need to use any mappings for the ID.  This filter also has the option for "None" to show entries that have no entity type.  Adds css that applies to the other filter to make it look nice.  Changes the location of the app.css import so that bootstrap styling does not override our local style.

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -450,19 +450,24 @@ em {
   visibility: visible;
 }
 
-.species-filter {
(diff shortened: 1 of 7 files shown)


azurebrd (approved):
Thanks @pjhale !

PR #271, opened 2024-05-20
adding complex and pathway TET UI sgd version and validating these two entities against SGD

sweng66:


azurebrd (approved):
Thanks @sweng66 !

sweng66:
Thanks @azurebrd

PR #273, opened 2024-05-21
Scrum 3678 Added facets for TET source fields

valearna:
- Added the new facets
- Added the ability to "show more" to all TET facets
- Removed extra calls to API to get initial facets
- Added a new redux state variable to avoid sending search requests until all default facet values are set (only mod_in_corpus_or_needs_review per logged in user for now)


sweng66 (approved):
Looks great! Thanks! @valearna

valearna:
thanks @sweng66 !

PR #275, opened 2024-05-22
adding multi filter to Topic and Entity columns

sweng66:
1. adding a generic multi filter that will shows up as a typeahead if there are more than 10 options; otherwise shows up as a list of checkboxes for the options.
2. creating TopicFilter and EntityFilter by calling the generic MultiFilter
3. applying the TopicFilter and EntityFilter  to the Topic and Entity Columns in the Topic Entity Table
4. Renaming original EntityFilter to EntityTypeFilter (it is for entity types, not for entities)

### DIFF
--- a/src/components/AgGrid/EntityFilter.jsx
+++ b/src/components/AgGrid/EntityFilter.jsx
@@ -1,7 +1,7 @@
 import { useGridFilter } from 'ag-grid-react';
 import React, { useCallback, useEffect, useState } from 'react';
-import {useSelector} from "react-redux";
-
+import { useSelector } from 'react-redux';
+import Select from 'react-select';
 
 export default ({ model, onModelChange }) => {
     const [closeFilter, setCloseFilter] = useState();
@@ -9,7 +9,7 @@ export default ({ model, onModelChange }) => {
     const allEntities = useSelector(state => state.biblio.allEntities);
     const doesFilterPass = useCallback((params) => {
         // doesFilterPass only gets called if the filter is active
-        return model.includes(params.data.entity_type_name);
+        return model.includes(params.data.entity_name);
     }, [model]);
 
     const afterGuiAttached = useCallback(({ hidePopup }) => {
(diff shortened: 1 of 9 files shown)


src/components/AgGrid/EntityFilter.jsx
```diff
@@ -1,15 +1,15 @@
 import { useGridFilter } from 'ag-grid-react';
 import React, { useCallback, useEffect, useState } from 'react';
-import {useSelector} from "react-redux";
-
+import { useSelector } from 'react-redux';
+import Select from 'react-select';
```
azurebrd, line 4: Is react-select already in the repo, or do you need that in the package.json ?  If it's new, I'll defer to Valerio and Paul about whether we need that, or should use something else (I have no idea)

sweng66: It is already in package.json.

sweng66: "react-select": "^5.8.0"

azurebrd (commented):
I still haven't worked with AgGrid, so I'll defer to other people.  In general looks good, just a question about whether we need react-select as an import in the package.json  and I'll defer to other people about that too.

src/components/AgGrid/EntityFilter.jsx
azurebrd: Ah, cool, thanks @sweng66 .  I didn't see it when I pulled main.

sweng66:
Thanks! @azurebrd Good question.

src/components/AgGrid/EntityFilter.jsx
azurebrd: https://github.com/alliance-genome/agr_literature_ui/blob/main/package.json

sweng66: You are right. For some reason, it is in my local version. I added the following to package.json.

"react-select": "^5.8.0"

Good catch! Thanks. @azurebrd

valearna (commented):
These changes look good to me. I tried using the UI and it looks great. The component is a little small. Can you make it taller so that there's no double scrolling in the select @sweng66 ?

sweng66:
@valearna I have fixed code to make the typeahead box taller.

sweng66:
@valearna I has fixed the code. Can you check again in dev3006? Thanks!

valearna (approved):
Thanks @sweng66 !

PR #277, opened 2024-05-22
do check for prof cur

ianlongden:


azurebrd (commented):
I haven't worked with AgGrid yet, deferring to other people

PR #280, opened 2024-05-23
adding code to validate systematic names for genes

sweng66:
also updated SGD version of TET creator to set the entity type based on the topic selected.


azurebrd (approved):
Great, thanks @sweng66 !

sweng66:
Thanks ! @azurebrd

PR #281, opened 2024-05-24
Sometimes this line resolves to bucket.docs_count.doc_count but bucke…

azurebrd:
…t.docs_count is undefined, and on 3002 that crashes the page.  I don't know why that doesn't happen on stage.  I think this fixes it without breaking anything, but I don't really know how search works, so it might be breaking something

### DIFF
--- a/src/components/search/Facets.js
+++ b/src/components/search/Facets.js
@@ -226,7 +226,7 @@ const Facet = ({facetsToInclude, renameFacets}) => {
                                                 </Col>
                                                 <Col>
                                                 <Badge variant="secondary">
-                                                    {['topics', 'confidence_levels', 'source_methods', 'source_evidence_assertions'].includes(key) ? bucket.docs_count.doc_count : bucket.doc_count}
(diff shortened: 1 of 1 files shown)


azurebrd:
@valearna @pjhale I'm tagging just the two of you, because I'm not sure if this fixes something, or if it wasn't really a problem why it wasn't a problem, or whether it might be introducing unexpected behaviour because I don't understand the underlying data structure.

sweng66:
@azurebrd I added this line for picking up reverse nested counting (counting docs instead of counting tags), but forgot to check if the reverse count exists. It is a great fix. Thanks!

valearna (approved):
The fix looks good, thanks @azurebrd !

azurebrd:
Great, thanks @sweng66 and @valearna !  I didn't realize you worked on search, Shuai, cool cool.  I also didn't know if there was some docs_count.doc_count vs doc_count based on those specific fields, or how the data was meant to be structured.

PR #283, opened 2024-05-24
fixing css styling for facets section to avoid wrapping and to make it more responsive

sweng66:


azurebrd (commented):
I'm not sure what this involves, so I'll defer to Valerio.  But I appreciate the link to 3006, I poked a bit, and wasn't sure what to look at .

sweng66:
Thanks for checking this out ! @azurebrd 
No problem. Let's look at it together next week.

The goal of this fix is to improve the wrapping issue of search facets. When the browser window is small, the facets on the stage have noticeable wrapping problems. In dev3006, the search facet section should perform better, at least in Chrome.

valearna:
The facets look better with these changes, but there are still some wrapping issues when the size is reduced. I'm not sure what we need to achieve here, maybe we should set a minimum width for the left menu. I would talk about this next week with the rest of the group.

sweng66:
Sounds good ! (talk about this as a group next week) @valearna

sweng66:
minWidth: "325px", // set minimum width for the facets section

I did set the min width. seems like this is still too small

azurebrd:
I think in general one should use rem or em, which are relative to font size, in case people are zooming in and out.

sweng66:
@azurebrd I am not familiar with rem or em. Can we have a team programming section next Tuesday to work it together?

azurebrd:
Yeah, that'd be great, @sweng66  !  

I think em makes things on a size relative to the font, and rem relative to some base browser or page font, I forget exactly, but we can look it up

sweng66:
Awesome!! @azurebrd

valearna (approved):
Everything looks good to me now. Thanks @sweng66 and @azurebrd

sweng66:
Thanks @valearna @azurebrd

PR #289, opened 2024-06-17
Scrum 4030

sweng66:
Add various dates to the breadcrumb display on ABC search page

### DIFF
--- a/src/components/search/BreadCrumbs.js
+++ b/src/components/search/BreadCrumbs.js
@@ -2,47 +2,126 @@ import React from 'react';
 import Container from "react-bootstrap/Container";
 import Row from "react-bootstrap/Row";
 import Col from "react-bootstrap/Col";
-import {useDispatch, useSelector} from "react-redux";
+import { useDispatch, useSelector } from "react-redux";
 import Button from "react-bootstrap/Button";
-import {RiCloseFill} from "react-icons/ri";
-import {removeFacetValue, resetFacetValues, removeExcludedFacetValue} from "../../actions/searchActions";
-import {RENAME_FACETS} from "./Facets";
+import { RiCloseFill } from "react-icons/ri";
+import {
+    removeFacetValue,
+    resetFacetValues,
+    removeExcludedFacetValue,
+    removeDatePubmedAdded,
+    removeDatePubmedModified,
+    removeDatePublished,
+    removeDateCreated,
+    searchReferences
+} from "../../actions/searchActions";
+import { RENAME_FACETS } from "./Facets";
 
 const BreadCrumbs = () => {
-
     const searchFacetsValues = useSelector(state => state.search.searchFacetsValues);
     const searchExcludedFacetsValues = useSelector(state => state.search.searchExcludedFacetsValues);
+    const datePubmedAdded = useSelector(state => state.search.datePubmedAdded);
+    const datePubmedModified = useSelector(state => state.search.datePubmedModified);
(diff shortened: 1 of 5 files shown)


azurebrd (commented):
I'll defer to other people on Search.  Thanks @sweng66 !

sweng66:
Thanks @azurebrd

src/components/search/BreadCrumbs.js
```diff
@@ -2,47 +2,126 @@ import React from 'react';
...
                     )}
-                    {Object.entries(searchExcludedFacetsValues).map(([facet, values]) =>
-                        <span key={facet + "_group"}>
-                            {values.map(value =>
-                                <span key={value + "_breadcrumb"}>
-                                    <Button variant="outline-danger">Exclude&nbsp;{(RENAME_FACETS.hasOwnProperty(facet) ? RENAME_FACETS[facet] : facet.replace('.keyword', '').replaceAll('_', ' ')) + ": " + value} &nbsp;
-                                        <RiCloseFill onClick={() => dispatch(removeExcludedFacetValue(facet, value))}/>
-                                    </Button>&nbsp;&nbsp;
-                                </span>)}
+                    {datePubmedModified && (
```
valearna, line 95: @sweng66 can you refactor this part and create a separate component to avoid repeating the same code multiple times?

src/components/search/BreadCrumbs.js
```diff
@@ -2,47 +2,126 @@ import React from 'react';
...
+                            <span key={facet + value + "_breadcrumb"}>
+                                <Button variant="outline-secondary">
+                                    {(RENAME_FACETS.hasOwnProperty(facet) ? RENAME_FACETS[facet] : facet.replace('.keyword', '').replaceAll('_', ' ')) + ": " + value} &nbsp;
+                                    <RiCloseFill onClick={() => handleRemoveFacet(facet, value)} />
+                                </Button>&nbsp;&nbsp;
+                            </span>
+                        )) : (
+                            <span key={facet + "_breadcrumb"}>
+                                <Button variant="outline-secondary">
+                                    {`${facet.replace('datePubmedAdded', 'Date Added to PubMed')
```
valearna, line 69: Here all dates are hardcoded. Is there a way to make the code more flexible and derive the names from the facets?

valearna (commented):
The PR looks good but I think it needs some refactoring to avoid duplication and hardcoded values

src/components/search/BreadCrumbs.js
sweng66: Will try to do so. Thanks !! @valearna

pjhale:
I would echo some of Valerio's concerns.  We could also change the breadcrumb text to something more consistent with the other breadcrumbs such as "Date Range: Published" or something similar.

sweng66:
Thanks both for code review and good suggestions @azurebrd @pjhale

src/components/search/BreadCrumbs.js
sweng66: fixed

src/components/search/BreadCrumbs.js
sweng66: fixed

sweng66:
updated code  @pjhale @valearna 

Please let me know if I missed anything.

valearna (approved):
Thanks @sweng66 , it looks good to me after the latest changes

sweng66:
Thanks @valearna

PR #292, opened 2024-06-24
Scrum 2670 Edit Mode

pjhale:


pjhale:
@azurebrd Made the merge and everything seems to be working but you might want to test and make sure I didn't clobber any code that you needed.

azurebrd (approved):
Cool !  Thanks @pjhale    I've loaded the branch, and see that the WB validation for creation works well, and played a little bit with loading with the edit button to the editor, but did not make any test patching.

We'll ask @cerivs for a story for editing the wb_validation.  Probably also the SGD validation ?  Or even SGD editing altogether ?

For the wb_validation load we can probably change TopicEntityTagActions.jsx in the line 92 else block to check for props.data.entity_id_validation === 'wb' + entity_type === 'gene'  and use the value 'use_wb' for changeFieldEntityAddTaxonSelect and pass the props.data.species with changeFieldEntityAddGeneralField to id taxonSelectWB .    Then for the PATCH in TopicEntityCreate.js we can copy the POST code from line 261.  I think that should do it, and hopefully it would just all work.

It's a bit odd that TopicEntityTagActions.jsx  line 88 sets editTag to null, but the biblioReducer.js line 60 initializes that as false.  Does it ever need false vs null to know that it's in an untouched state versus selected and de-selected ?

There's some formatting changes that aren't a big deal.  TopicEntityCreate.js  
line 120 remove unwanted spaces, good.
line 259 adding spaces seems odd, but maybe it's to line up the = assignment ?
line 393 a commented out selected value, probably not needed, but if you want it on purpose that's fine.

Thanks !

azurebrd:
cool cool, thanks @pjhale

PR #299, opened 2024-06-28
When loading to Edit from props.data check for entity_id_validation b…

azurebrd:
…eing WB to load taxonSelect and taxonSelectWB, otherwise load taxonSelect.  Generalize with changeFieldEntityAddGeneralField instead of changeFieldEntityAddTaxonSelect.  Patching data should also use WB values like Post does.


valearna (approved):

azurebrd:
Thanks @valearna !

PR #300, opened 2024-06-29
Using converted name field for source evidence assertion

valearna:


sweng66 (approved):
Looks good! Thanks @valearna

PR #304, opened 2024-07-11
Recreate items object from initItems if value in cookie does not contain all the needed items

valearna:


sweng66 (approved):
Thanks for fixing this. @valearna

valearna:
Thank you @sweng66 !

PR #307, opened 2024-07-22
Fixing Bug

pjhale:
The props are now updated on rowUpdate so a further update has the correct data.


azurebrd (approved):
Great, thanks @pjhale !

PR #314, opened 2024-08-12
fixed code to add 'New row' and 'Submit All' buttons to SGD version of TET UI

sweng66:
to allow curators to add multiple topics at the same time
also cleaned up the other MOD version of TET interface


valearna (approved):

sweng66:
Thanks @valearna

PR #315, opened 2024-08-13
fixed "Create reference UI" to generate SGDID directly through API when…

sweng66:
… adding a new paper to SGD corpus


valearna (approved):

sweng66:
Thanks @valearna

PR #322, opened 2024-08-26
fixed to not reset the cloned rows when the topic changes

sweng66:


azurebrd (approved):
I don't really get how this works in the context of everything else, but it makes sense.

sweng66:
thanks @azurebrd

PR #323, opened 2024-08-29
add more fields to TET table column options.  clean up some warnings.

azurebrd:
@sweng66 @pjhale There's some warnings, and I fixed a few that seemed obvious.  There's still a few more though

  Line 51:6:    React Hook useEffect has a missing dependency: 'dispatch'. Either include it or remove the dependency array        react-hooks/exhaustive-deps
  Line 79:6:    React Hook useEffect has a missing dependency: 'fetchTableData'. Either include it or remove the dependency array  react-hooks/exhaustive-deps
  Line 348:9:   'nameFormatter' is assigned a value but never used                                                                 no-unused-vars
  Line 425:19:  'setColDefs' is assigned a value but never used   

348 ->  can we just get rid of that ?  It doesn't seem to be used anywhere
425 ->   const [colDefs, setColDefs] = useState(cols);
   colDefs is only used in one place
            columnDefs={colDefs}
   can we just use cols there ?  Because we never use setColDefs anywhere.

51 and 79 I'm not sure about.  I feel like we've run against that before, but I don't recall what we decided for those cases.  Maybe we add them to the array, or we add an es-lint comment to prevent the warning ?

### DIFF
--- a/src/components/biblio/topic_entity_tag/TopicEntityTable.js
+++ b/src/components/biblio/topic_entity_tag/TopicEntityTable.js
@@ -31,9 +31,6 @@ const TopicEntityTable = () => {
   const biblioUpdatingEntityAdd = useSelector(state => state.biblio.biblioUpdatingEntityAdd);
   const referenceCurie = useSelector(state => state.biblio.referenceCurie);
   const [isLoadingData, setIsLoadingData] = useState(false);
-  const ecoToName = {
-    'ECO:0000302': 'author statement used in manual assertion'
-  };
   const [selectedCurie, setSelectedCurie] = useState(null);
   const [showModal, setShowModal] = useState(false);
   const [fullNote, setFullNote] = useState('');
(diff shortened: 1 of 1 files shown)


azurebrd:
Sounds good @sweng66 , I've added those changes.  Do you think we can do the first 2 changes I mentioned ?

sweng66:
I think so @azurebrd

sweng66:
also we can get rid of "ecoToName"? since we are getting rid of "nameFormatter'"

sweng66:
Do we need to move "fetchTableData" inside useEffect?

useEffect(() => {
    const fetchTableData = async () => {
       // code here
  };

  fetchTableData();
}, [topicEntityTags, biblioUpdatingEntityAdd]);

sweng66 (approved):
Looks good! I see the new columns in the table and in the Hide/Show Columns list.Thanks @azurebrd

azurebrd:
Thanks @sweng66 !  I don't know about moving the const definition of fetchTableData.  I don't use useEffect much, and I didn't write that.

Sounds good, I got rid of ecoToName and nameFormatter.

I guess we're leaving colDefs for now ?   I don't know, I also don't use useState much.

sweng66:
Sounds good ( leaving colDefs for now)! Thanks @azurebrd

azurebrd:
Thanks @sweng66 !

PR #324, opened 2024-08-31
Display cross references like resources_for_curation with pipe separa…

azurebrd:
…tion in one cell instead of across multiple rows.  Add that to Biblio Display and general section for TET and WF


sweng66 (approved):
Looks good! Thanks @azurebrd

azurebrd:
Thanks @sweng66 !

PR #326, opened 2024-09-03
Scrum 4390 - call TET API only once (and not an infinite number of times)

valearna:
Also, avoid calling the API twice when loading the data for the first time


sweng66 (approved):
Thanks @valearna

PR #335, opened 2024-09-30
adding code to show 'Add this paper to corpus' if the paper is not in logged-in user's mod

sweng66:
and hide the file upload section if the paper is outside logged-mod.


azurebrd (approved):
Thanks @sweng66 !

sweng66:
Thanks @azurebrd

PR #339, opened 2024-10-11
Fixing Initial State for TET Variables

pjhale:
In the reducer we are setting filteredTags and editTag to null so that when a new page is opened the TET table has no variables/filters set.


azurebrd (approved):
Thanks @pjhale !

PR #340, opened 2024-10-15
updating sort page to work for single reference

sweng66:
* adding paper into log-in user's mod corpus
* adding paper into log-in user's mod corpus plus directing curators to TET page
* discarding paper from log-in user's mod
* adding "Update Sorting" button on the top (along with the one at the bottom) to discard/add multiple papers at the same time
* also fixed the search code to handle tester properly (for missing PDF icon issue in search result page)

### DIFF
--- a/src/actions/searchActions.js
+++ b/src/actions/searchActions.js
@@ -100,7 +100,7 @@ const getSearchParams = (state) => {
     query_fields: state.search.query_fields,
     sort_by_published_date_order: state.search.sortByPublishedDate,
     partial_match: state.search.partialMatch,
-    mod_abbreviation: state.isLogged.oktaMod
+    mod_abbreviation: state.isLogged.testerMod !== 'No' ? state.isLogged.testerMod : state.isLogged.oktaMod
   }
 
   const data = state.search.searchFacetsValues;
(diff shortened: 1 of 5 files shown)


src/actions/searchActions.js
```diff
@@ -100,7 +100,7 @@ const getSearchParams = (state) => {
     query_fields: state.search.query_fields,
     sort_by_published_date_order: state.search.sortByPublishedDate,
     partial_match: state.search.partialMatch,
-    mod_abbreviation: state.isLogged.oktaMod
+    mod_abbreviation: state.isLogged.testerMod !== 'No' ? state.isLogged.testerMod : state.isLogged.oktaMod
```
azurebrd, line 103: Does this handle the case where there is no okta login and no tester value ?  (I don't know if mod_abbreviation is required for this)

azurebrd (approved):
I honestly lost track of what's supposed to happen, but the code looks good, thanks @sweng66

src/actions/searchActions.js
sweng66: Good question! In this special case, mod_abbreviation is null, the main PDF identifying code will return nothing so no main PDF icon will show up on the search result page.

Currently, the PDF icon for a paper with a main PDF uploaded by SGD, does not appear on the search results page when viewed by a non-SGD curator, even if they are logged in as an SGD tester. With this update, we will set mod_abbreviation = 'SGD' when a non-SGD curator is logged in as an SGD tester, making sure the main PDF icon is visible for a tester as well.

sweng66:
Thanks @azurebrd for looking over the code.

This fix (adding "Inside & Open", "Inside", "Outside" buttons) is mainly for SGD paper sorting/triaging. One paper at a time.

The "Inside & Open": adding the given paper to mod corpus and redirecting curator to TET page.
"Inside": adding the paper into mod corpus
"Outside": removing the paper out of mod corpus

Sorry this PR is for two stories so it is a bit messing (too lazy to create a new one).

src/actions/searchActions.js
azurebrd: Oh... cool cool, thanks !

azurebrd:
Got, it thanks @sweng66 !

PR #343, opened 2024-10-31
adding code to catch / display the response.data.detail from 'delete' PDF API

sweng66:
and to give curators the options to choose to keep OR delete manual tet tags after deleting a main PDF. Also refactored the FileEditor


azurebrd (approved):
Thanks @sweng66 !

sweng66:
Thanks @azurebrd

PR #350, opened 2024-11-15
To avoid error 'Cannot update during an existing state transition (su…

azurebrd:
…ch as within render). Render methods should be a pure function of props and state.'  Change pushHistory to not happen conditionally on createRedirectToBiblio, but rather a useEffect based on createRedirectToBiblio.

### DIFF
--- a/src/components/Create.js
+++ b/src/components/Create.js
@@ -253,6 +253,13 @@ const Create = () => {
     dispatch(setReferenceCurie(referenceCurie));
     history.push("/Biblio/?action=editor&referenceCurie=" + referenceCurie);
   }
+
+  useEffect(() => {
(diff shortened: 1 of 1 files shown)


sweng66 (approved):
Looks great! Thanks @azurebrd

valearna (approved):
Not sure what this does but looks good

azurebrd:
Thanks @sweng66 !

@valearna the code that was there before was conditional on createRedirectToBiblio but there was a browser warning 'Render methods should be a pure function of props and state.'  And putting it in a useEffect let it work.  I guess, whenever we got that history.push into the codebase, we thought it was working, but didn't check that it wasn't.

PR #352, opened 2024-11-18
fixed the TET data loading spinning issue

sweng66:


valearna (approved):
Not exactly sure how this works, but looks good. Thanks @sweng66 !

sweng66:
Thanks @valearna 

The newly updated useEffect now directly reacts to changes in biblioUpdatingEntityAdd. The fetchTableData is called whenever a new tag is added or updated. Removing the dependency on topicEntityTags magically fixed the issue

PR #354, opened 2024-11-25
New button to set file unavailable.  Both this button and previous bu…

azurebrd:
…tton in progress now should if the mod's file status is file needed or in progress.  I think it's weird that if in progress you can click to make in progress, but it's what the story says.  refactor handleFileUnavailableClick to generally handleWorkflowTransition and return axios promise for await, when needing to step through two workflows in sequence from separate api calls.  if setting to file unavailable from file needed, first go through in progress, because API requires that


sweng66 (approved):
So we are transitioning from "file needed" to "file upload in progress", then to "file unavailable" in the UI instead of letting API handling this. That is good. Thanks @azurebrd

azurebrd:
Thanks @sweng66 !

PR #355, opened 2024-11-26
story ambiguous if in progress button should show for file needed and…

azurebrd:
… in progress or unavailable, but unavailable makes more sense


sweng66 (approved):
Thanks @azurebrd

azurebrd:
Thanks @sweng66 !

PR #363, opened 2024-12-13
adding check and displaying a warning message if another curator is adding or done adding TETs

sweng66:


azurebrd (approved):
Thanks @sweng66 !

PR #365, opened 2024-12-16
fixed to add a close button for warning message

sweng66:


sweng66:
@azurebrd Ceri mentioned we can deploy whatever is ready to prod first. Then we can deploy this fix (suggested by Ceri) tomorrow as needed.

valearna (approved):

PR #370, opened 2025-01-23
updated to replace A-team API calls with newly set up ABC API calls

sweng66:
Also asked ChatGPT to clean up TopicEntityCreateSGD.js...

### DIFF
--- a/src/actions/biblioActions.js
+++ b/src/actions/biblioActions.js
@@ -437,7 +437,9 @@ export const wb_entity_validation = (dispatch, entityType, entityInputList, call
 
 export const abc_entity_validation = (dispatch, entityType, entityInputList, taxon, callback) => {
 
-    const url = restUrl + "/topic_entity_tag/entity_validation/" + taxon + "/" + entityType + '/' + entityInputList.join('|').replace(/ /g, '+');
+    const entityListStr = entityInputList.join('|');
+    const encodedEntityList = encodeURIComponent(entityListStr);
+    const url = `${restUrl}/topic_entity_tag/entity_validation/${taxon}/${entityType}/${encodedEntityList}`;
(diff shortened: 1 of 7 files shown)


src/actions/biblioActions.js
```diff
@@ -438,58 +435,47 @@ export const wb_entity_validation = (dispatch, entityType, entityInputList, call
...
-    let otherName = entityType === "construct" ? entityResult['constructFullName']?.displayText?.toLowerCase() ?? "" : "";
-    if (primaryId && name) {
-      if (entityResult.obsolete === true) {
-        obsoleteMap[primaryId.toLowerCase()] = primaryId;
-        obsoleteMap[name] = primaryId;
-        if (systematicName) {
-          obsoleteMap[systematicName] = primaryId;
+    const entityListStr = entityInputList.join('|');
+    const encodedEntityList = encodeURIComponent(entityListStr);
+    const url = `${restUrl}/topic_entity_tag/entity_validation/${taxon}/${entityType}/${encodedEntityList}`;
```
valearna, line 442: Is this GET a potential limit on the number of entities we can pass to the API? If so we could make it a POST and modify the API

valearna (approved):
All changes look good to me. I'm only concerned that GET requests may limit the amount of data we can send to the API

src/actions/biblioActions.js
sweng66: Good point. However, this is only for entity validation. I’m assuming curators won’t post a huge list to the entity box, so using the GET method should be sufficient. What do you think?

valearna: Oh, good point, so the mapping for TETs is done in the backend directly, right? Then this should be fine. I don't know how many entities curators can enter in the autocomplete

PR #376, opened 2025-02-07
updated search Facets to display Workflow tags in different sub-categories (File workflow, Entity extraction, etc)

sweng66:


valearna (approved):
Not sure if this includes the "show more" buttons but looks good

sweng66:
Thanks @valearna

sweng66:
Since "show more" is for each subcategory, I don't think we have more than 10 WF tags under each subcategory.

PR #377, opened 2025-02-11
adding code to make sub-categories expandable / closable  

sweng66:
and by default, they are closed


pjhale:
@sweng66 Was this supposed to apply to all facets (the collapsable part panel) or just WF tags?  If you intended it to be all facets then its working properly but it wasn't clear in the ticket if that was the case.

pjhale (approved):
I'll approve but if you didn't intend on this applying to all facets then you probably need to change it.

sweng66:
@cerivs agreed that we should apply this (the collapsable part panel) to all facets so we are good. Thanks for checking out the code. @pjhale

PR #381, opened 2025-02-19
Updated code to display the workflow tag names instead of ATP IDs in the navigational bar

sweng66:


azurebrd (approved):
Thanks @sweng66

sweng66:
Thanks @azurebrd

PR #382, opened 2025-02-19
seems at some point the id of the html element had a space and now it…

azurebrd:
… has a hyphen, so the reducer needs to split on that instead


sweng66 (approved):
Thanks @azurebrd

azurebrd:
Thanks @sweng66 !

PR #384, opened 2025-02-24
Update BiblioEdit and Sort UI to catch api error and display a warning when removing a paper out of corpus 

sweng66:
when the paper has manually added topic/entity tags


azurebrd (approved):
Thanks @sweng66 !

PR #390, opened 2025-03-17
WorkflowStatTableCounters takes in a third column type of date_range,…

azurebrd:
… which passes a date frequency to the API, and for a given mod returns time slices for that time period.  Adds this table to the top of each mod report.


sweng66 (approved):
Looks good! Thanks @azurebrd

azurebrd:
Thanks @sweng66 !

PR #401, opened 2025-04-08
adding code to show a tool tip that displays the source description when the source method is moused over

sweng66:


pjhale (approved):
Looks good.

sweng66:
Thanks @pjhale

PR #409, opened 2025-05-01
curators wanted resync table and download buttons to always line up v…

azurebrd:
…ertically, so now they're always as far to the right as possible


sweng66 (approved):

azurebrd:
Thanks @sweng66 !

PR #414, opened 2025-05-13
upgrading axios

sweng66:


azurebrd (approved):
Thanks @sweng66 .  I don't really know the difference between ^ ~ and >=  I could imagine that >= could eventually upgrade us into something that we're not expecting ?  But I really don't know.  If any of the more systems people have thoughts about it, that might be good.  Seems good to me if it's working though.

PR #415, opened 2025-05-14
note should be editable with a agLargeTextCellEditor but only if ther…

azurebrd:
…e's a valid curation status already in the database.  update API to work off of general parameter passed in, but it must be curation_status or note.


sweng66 (approved):
Looks good! Thanks @azurebrd

PR #419, opened 2025-05-19
Scrum 5074

pjhale:
Reopening this branch... added a date converter and the papers here are now clickable with a link that goes to the biblio page.  Also added padding between the two tables.

### DIFF
--- a/src/components/Reports.js
+++ b/src/components/Reports.js
@@ -31,6 +31,14 @@ const file_upload_name_mapping = {
     'file upload in progress': 'in progress'
 }
 
+const convertDate = (date) => {
+    if (date){
+        return date.replace(/(\d{4})(\d{2})(\d{2})/, "$1-$2-$3")
+    } else {
+        return null
+    }
+}
+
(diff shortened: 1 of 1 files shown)


src/components/Reports.js
```diff
@@ -684,11 +692,20 @@ const QCReportRetractedPapers = ({modSection}) => {
             }
         };
         if (qcReportRedactedPapers['date-produced'] === null) {fetchData(); }
-    }, [qcReportRedactedPapers]);
+    }, [qcReportRedactedPapers, dispatch]);
```
valearna, line 695: not sure this is required here. In the past we had to remove dispatch from useEffects to avoid constantly reloading pages. Can you check if this is causing the useEffect to constantly reload @pjhale ?

valearna (approved):

src/components/Reports.js
pjhale: Just added some logs... it is not causing the use effect to trigger.

valearna: sounds good, thanks

PR #423, opened 2025-05-22
curationWholePaperData is now an array with a single object, instead …

azurebrd:
…of the object, because we're using an aggrid table to edit.  replace the topic_name to Whole Paper for display.  popupParent is document.body because the note editor is bigger than the aggrid table.  don't show headers, it's just one row.  comment out CurationStatusWholePaperTextarea but I want it in main in case they want it back for some reason in the future, will remove once passes uat.  when changing the whole paper curation status into ATP:0000237 trigger setDataTopicsInProgress to change all the topics that have data but no curation status to have that value


sweng66 (approved):
Thanks @azurebrd

azurebrd:
Thanks @sweng66 !

PR #424, opened 2025-05-25
Show placeholder if curation status is value, and note is empty, othe…

azurebrd:
…rwise show blank


azurebrd:
Thanks @sweng66 !

PR #427, opened 2025-06-06
adding Manual Indexing and Community Curation table to Workflow edit

sweng66:


src/components/biblio/BiblioWorkflow.js
```diff
@@ -564,11 +639,126 @@ const BiblioWorkflow = () => {
...
+    } catch (error) {
+      const errorMessage = `API error: ${error.message}`;
+      setApiErrorMessage(errorMessage);
+      setShowApiErrorModal(true);
+      console.error('Error updating workflow tag:', error);
+    }
+  }, [referenceCurie, accessLevel, accessToken, fetchIndexingWorkflowOverview]);
+
+  // Helper function for POST request
+  const postWorkflowTag = async (data) => {
```
azurebrd, line 729: Oh, nice.  If you wanted to, we could make these higher level so that other places that update WorkflowTag use them, but I'm fine with them being a separate thing.  I think at this point there's only a couple of other places that do this in this code

PR #435, opened 2025-07-02
Warning row for curators not to create and delete authors at the same…

azurebrd:
… time.  The UI cannot handle gaps in the author list based on the  db author order from the API.  When a delete is sent to the API, it triggers comparing the author orders to what they should be, but the newly created author does not have an author_id from the database yet, so it cannot update its order to flatten the order gap.

 An API endpoint to flatten author orders would fix it, replace the  if (hasAuthorDeletion) { survivingAuthors  code.
 UI is not meant to handle gaps in order, it could be reworked to.


valearna (approved):

azurebrd:
Thanks @valearna !

PR #436, opened 2025-07-03
Additional message for curators to change browser settings to avoid a…

azurebrd:
…utomated okta logout


sweng66 (approved):
Thanks @azurebrd

PR #445, opened 2025-07-11
Confidence Facet

pjhale:
Initial commit.  This is mostly working but still need to filter out the confidence level from the tet facet stuff.


pjhale:
Should be ready for review.

PR #451, opened 2025-08-02
Scrum 5195

azurebrd:


azurebrd:
Claude Code "Critical Logic Error in getDataNoveltyAtpArray"  can be ignored.  Curators were clear at standup that it should work this way.  Create allows 2 values.  Edit only allows one value.

sweng66 (approved):
Looks good! Thanks @azurebrd

azurebrd:
Thanks @sweng66 !

PR #452, opened 2025-08-05
if an xref.is_obsolete is true, create a red obsolete span and displa…

azurebrd:
…y that before the xref


sweng66 (approved):
Looks good! Thanks @azurebrd

PR #454, opened 2025-08-07
Fix for Long Notes

pjhale:
We use auto column widths for this table.  But some notes are super long.  This change adds a listener to column resized.  We look for events that are auto and if the notes field width is more than 300 wide.


azurebrd (approved):
Thanks @pjhale !

PR #457, opened 2025-08-11
fixed code to get / patch indexing priority data with two new endpoints

sweng66:
- getting and updating the data in the new table 'indexing_priority'. Also fixed to display the confidence score if it is available


azurebrd (commented):
Thanks @sweng66 !

PR #458, opened 2025-08-11
SCRUM-5277 Automatic removal of NOT selections in confidence level.

pjhale:
When there are no topics selected the UI should automatically remove any NOT selections in confidence level as they do not make sense.  This checks in the facet and breadcrumbs every time a user removes all topics and then removes all NOT confidence level selections.  This does not remove positive selections in this facet.

Also removed a searchReferences() call that appears redundant.  These are in other functions here but I didn't touch them.


sweng66 (approved):
Thanks @pjhale

PR #459, opened 2025-08-11
suffix should always be lowercased to make case insensitive check to …

azurebrd:
…set pdfType.  pdfType can only be from fixed list, otherwise use pdf instead of null.  htm should still map to html.  temp suffix always returns pdf and temp.


azurebrd:
Thanks @sweng66 !

PR #461, opened 2025-08-27
account for facets of source_evidence_assertions by fetching their de…

azurebrd:
…scriptions from ateam based on their eco or atp code, and use the same Tooltip overlay as source_methods


sweng66 (approved):
Looks good! Thanks @azurebrd

PR #467, opened 2025-09-09
Add a button for In Progress, beside the existing one for Done.  Use …

azurebrd:
…different endpoint /workflow_tag/transition_to_workflow_status for updating.  After successful update, fetch referenceJsonLive from db to update button state


sweng66 (approved):
Thanks @azurebrd

PR #477, opened 2025-09-30
new WorkflowTagService to delete a workflow_tag.  GeneralizedDropdown…

azurebrd:
…Renderer allows a blank option when workflow_tag as well as previous curation_status.  when clearing a workflow_tag from an existing value to no value, delete the workflow_tag


azurebrd:
Thanks @sweng66 !

PR #478, opened 2025-10-01
Add Headers to the Manual Indexing and Community Curation AgGrid tabl…

azurebrd:
…e, and change the header names to match the story, but not changing the field names, since those better match where the data comes from the API.  Also setting height to autoHeight which fits, but doesn't look great


sweng66 (approved):

valearna (approved):
Thanks @azurebrd

azurebrd:
Thanks @sweng66 and @valearna !

PR #579, opened 2026-05-14
fix(biblio): prevent DateRangePicker crash on unparseable publication…

azurebrd:
… dates

BiblioDateComponent passed any non-null date_published_start/end value straight into new Date(), so a malformed string (e.g. "nullT00:00:00") became an Invalid Date that react-date-picker's DateInput threw on inside a mount effect, blocking the editor from loading the reference. Coerce unparseable values to null and surface the raw stored string inline so curators can see what was stored and re-select a range.


valearna (approved):
cool, thanks @azurebrd !

azurebrd:
Thanks @valearna !

PR #580, opened 2026-05-16
link image permission name to permission URL

sweng66:
In the file management tab, the image_permission_name is now displayed as a clickable link to the permission_url when available. This allows users to easily access the publisher's permission page.


valearna (approved):

sweng66:
Thanks @valearna

PR #584, opened 2026-05-20
feat(SCRUM-4812): add Models sub-tab to Reports for /ml_model/all

azurebrd:
Surfaces the new /ml_model/all endpoint as a "Models" sub-tab inside the All tab and inside each MOD tab that has at least one model. Data is fetched once and cached in Redux so switching MOD tabs does not re-hit the API. Table reuses TET's AG Grid patterns (column filters, sorting, click-to-modal for long-text description/parameters cells, TSV download dropdown) and shares a single person_setting row via componentName="models_table" across every instance.


azurebrd:
@claude review again

sweng66 (approved):
Thanks @azurebrd

azurebrd:
Thanks @sweng66 !

PR #586, opened 2026-05-22
adding workflow diagram

sweng66:


valearna (dismissed):
Looks great, thanks @sweng66 !

sweng66:
@valearna  Thanks. Tried to get rid of hardcoded ATP IDs for inter-workflow connections (eg, from text connection to entity extraction etc from actions info), but failed so will just leave as is for now.  Can you check again? no rush

valearna (approved):

sweng66:
Thanks @valearna

PR #588, opened 2026-05-26
fix(SCRUM-6135): poll version.json periodically so banner appears in Firefox

valearna:
## Summary
- `useVersionCheck` only ran on mount and on `visibilitychange`. Firefox doesn't fire `visibilitychange` while a tab stays in the foreground, so the "new version available" banner never appeared unless the user switched tabs. Chrome was masking the bug by firing visibility events more aggressively.
- Added a 5-minute `setInterval` inside `useVersionCheck` (same pattern as `DebeziumStatusAlert.js:36`) so the check runs even when the tab stays focused.
- The existing `MIN_CHECK_INTERVAL_MS` (1 min) throttle and the `updateAvailableRef` short-circuit keep request volume bounded and stop polling once an update is detected.

Jira: [SCRUM-6135](https://agr-jira.atlassian.net/browse/SCRUM-6135)

## Test plan
- [ ] Load the UI in Firefox, leave the tab focused, deploy a new build (or manually change `/version.json`'s `buildId`), and confirm the banner appears within ~5 minutes without any tab switch.
- [ ] Confirm Chrome behavior is unchanged (banner still appears).
- [ ] Confirm only one `/version.json` request fires per minute even with focus/visibility churn (throttle still effective).
- [ ] Confirm no further requests fire after the banner appears (short-circuit still effective).

[SCRUM-6135]: https://agr-jira.atlassian.net/browse/SCRUM-6135?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/hooks/useVersionCheck.js
+++ b/src/hooks/useVersionCheck.js
@@ -2,10 +2,12 @@ import { useEffect, useRef, useState, useCallback } from 'react';
 
 /**
  * Hook that checks for new application versions by polling /version.json.
- * Checks on initial mount and whenever the tab becomes visible.
- * Returns { updateAvailable: boolean } when a newer build is detected.
+ * Checks on initial mount, on a periodic interval, and whenever the tab
+ * becomes visible. Returns { updateAvailable: boolean } when a newer build
+ * is detected.
  */
 const MIN_CHECK_INTERVAL_MS = 60_000; // Don't check more than once per minute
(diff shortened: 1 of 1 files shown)


sweng66 (approved):
Thanks @valearna

valearna:
thanks @sweng66 !

PR #589, opened 2026-05-27
feat(SCRUM-523): group PDB and GEO references under new datasets section

ianlongden:
Filter PDB and GEO cross_references out of the cross_references row and render them in a new "datasets" row immediately below, in both the biblio display and editor views. The underlying data layout is unchanged -- entries still live in the cross_references array -- so API submission, revert, and delete paths are unaffected.


ianlongden:
I started stage with this branch, so can be viewed too. An example is https://stage-literature.alliancegenome.org/Biblio/?action=display&referenceCurie=AGRKB:101000000342002

valearna (approved):

PR #591, opened 2026-05-30
Scrum 5716

azurebrd:


sweng66 (approved):
Thanks for fixing all these. @azurebrd 

in src/actions/createActions.js‎, line 276-277:

In theory, a successful response should always have valid res.data.curie.. so no need to add a Null Checks on Response Data, but do you think it will still be nice to add one?.. I dont know? 

  
if (isSuccess(res.status)) {                                                                                              
         const newResourceId = res.data?.curie;                                                                                  
         if (!newResourceId) {                                                                                                   
                 console.error('Unexpected response shape:', res.data);                                                                
        }                                                                                                                       
        // proceed...                                                                                                           
  }

azurebrd:
@sweng66 

CC says

no, but a soft "no"

  The backend now declares response_model=ResourceSchemaShow on POST /resource/add/, and curie is a non-optional field on that schema. FastAPI will
  refuse to send a response that doesn't include it (it'd surface as a server-side ResponseValidationError → 500). So res.data.curie is contractually
  guaranteed for 2xx responses.

  The reviewer's defensive version:
  const newResourceId = res.data?.curie;
  if (!newResourceId) {
    console.error('Unexpected response shape:', res.data);
  }
  costs almost nothing — three lines — and gains a console message if the backend ever regresses. But it also masks a backend regression rather than
  letting it surface loudly. Today, if the contract broke, you'd see 'Resource created successfully: undefined' appear in Redux state, which is at
  least immediately visible. With the defensive check, you'd see the same undefined plus a console log nobody reads.

  I'd say: don't add it, because:
  - it's inconsistent with how the other response?.[subField] ?? null reads in the codebase work (those propagate null to Redux silently; they don't
  console.error),
  - Pydantic's response_model is the contract enforcement point — trusting it is the whole reason we moved to strict shapes,
  - the reviewer's ?. + log pattern is a defensive idiom that only pays off if backend contracts are unreliable, which Phase 1 was specifically meant
  to fix.

  If the reviewer feels strongly, the cost of adding it is genuinely tiny — under a line per call site. I just don't think it earns its keep.

PR #593, opened 2026-06-01
feat(search): hide GEO and PDB cross references on search results

ianlongden:
Filter dataset xrefs (GEO, PDB) out of the per-result xref list so they no longer clutter the search page. Reuses datasetXrefPrefixes from BiblioEditor as the canonical prefix list.


sweng66 (approved):
Thanks @ianlongden

PR #594, opened 2026-06-01
 add Curation Classification Tags facet categories

sweng66:
New facets under "Curation Classification Tags":
- Predicted indexing priority
- Indexing priority
- Manual indexing curation tag


valearna (approved):

sweng66:
Thanks @valearna

PR #595, opened 2026-06-02
left-align facet category headers to fix indentation

sweng66:
The "Curation Classification Tags" category title wrapped to two lines and appeared indented because Bootstrap's default centered button text pushed the arrow icon and wrapped text inward. Left-aligning the category button text keeps all category headers flush-left.


valearna (approved):

sweng66:
Thanks @valearna

PR #596, opened 2026-06-03
relabel predicted curation tag and keep category header on one line

sweng66:
- Display "no genetic data" as "predicted no genetic data" in the Manual indexing curation tag facet, since the value comes from the predicted curation tag column and is a computed prediction, not a curator-added tag. Added RENAME_FACET_VALUES map for per-facet bucket value relabeling.

- Prevent the "Curation Classification Tags" category header from wrapping to two lines by using whiteSpace: nowrap with a slightly smaller font/padding so the longest header fits within the facet sidebar.

### DIFF
--- a/src/components/search/Facets.js
+++ b/src/components/search/Facets.js
@@ -104,6 +104,18 @@ export const RENAME_FACETS = {
 }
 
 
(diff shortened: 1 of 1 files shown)


valearna (approved):

sweng66:
Thanks @valearna

PR #599, opened 2026-06-05
fix(SCRUM-6071): mount banner directory and serve via nginx alias for…

azurebrd:
… live edits

Mount a directory (BANNER_DIR, default ./banner) at /var/www/banner-config instead of bind-mounting the single banner.json file, and serve /banner.json via nginx alias. Because the path is resolved per request, host edits -- including vim/sed save-by-rename -- are picked up live with no container restart, matching the public/banner.json experience under npm start.

BANNER_DIR points at a stable, server-managed location (e.g. /usr/share/agr_ui_files) set in the env file, so the banner content survives the disposable Jenkins workspace across redeploys.


sweng66 (approved):
Thanks @azurebrd

PR #600, opened 2026-06-08
feat(SCRUM-6046): add graphical 3-section layout preference for Bibli…

azurebrd:
…oEditor

Add a "Layout" modal that lets curators arrange the editor's three sections (Vital / Authors / Additional Info) in a 2D grid via drag-and-drop, saved per-user as named layouts.

- Sections are derived as contiguous slices of fieldsOrdered (split at authors); rows are routed into section buckets via a push shim so the existing render branches are untouched.
- The editor renders the sections through a CSS Grid mapped from the saved react-grid-layout coordinates (layoutToCssGrid). Side-by-side arrangements switch the editor to a full-width container with an adaptive per-column minimum width chosen by how many sections share a row (columnFloorForLayout: ~120px for two-up, ~280px for three-up), scrolling the page when needed.
- BiblioLayoutPreferenceModal provides the drag canvas plus named-layout management (create / load / save / rename / set-default / delete) persisted via usePersonSettings under the biblio_editor_layout namespace.
- Adds react-grid-layout dependency; new biblioEditorSections.{js,css,test.js}.

CC alphabetized the package.json  the only addition is "react-grid-layout": "^1.5.3",


sweng66 (approved):
Thanks @azurebrd

azurebrd:
Thanks @sweng66 !

PR #602, opened 2026-06-10
add include/exclude toggle to source method and SEA facets

sweng66:
Route source method and source evidence assertion exclusions through the nested-TET negation path (tet_facets_negative_values) so curators can exclude manually-added tags and find only pipeline-added tags.


sweng66:
@claude Check again?

azurebrd (approved):

sweng66:
Thanks @azurebrd

PR #604, opened 2026-06-15
show SEA label instead of curie in excluded breadcrumb

sweng66:
The breadcrumb read the human-readable label from bucket.name, which the backend only attaches to buckets present in the current aggregation. When a source evidence assertion value is excluded, its bucket drops out of the aggregation, so the breadcrumb fell back to the raw curie (e.g. ECO:0006155).

Translate the ECO/ATP curie to its short label via the existing evidenceAssertionName map when no live bucket name is available.


sweng66:
@claude Check again?

valearna (approved):

sweng66:
Thanks @valearna

PR #607, opened 2026-06-17
bind confidence score to the selected tag in single-tag mode

sweng66:
In single-tag mode the confidence slider was pushed as its own object in tet_facets_values, which the backend turns into an independent nested query. That let the confidence range be satisfied by any tag on a reference rather than the tag matching the selected topic. Merge confidence_score into the combined nested object so both conditions must hold on the same tag.


azurebrd (approved):

sweng66:
Thanks @azurebrd

PR #610, opened 2026-06-24
honor Exclude NEG / confidence in Topic grid + speed it up

sweng66:
Two things: (1) the Topic grid showed negative/low-confidence data even when the search excluded it, and (2) the grid was slow to load.

## Fixes — Exclude NEG / confidence (display correctness)
The grid fetches TETs independently of the search query, so it never applied the
search's confidence filters. Now it filters its TETs to match the search:
- Drops `confidence_level == 'NEG'` when the **Exclude NEG** chip is set
  (= the "confidence score ≤ 0.5" bucket, matching the backend `must_not`).
- Drops any TET whose numeric `confidence_score` is outside the **slider** range.
- Filtering is applied once at the row source, so every per-topic column
  (Sources / Data / Conf / Note) stays consistent.
- Also removed the `applyToSingleTag` gate in `searchActions` so **Exclude NEG**
  filters the search in **multi-tag** mode too — the backend already applies it
  per-tag, so this just stops silently dropping the filter.

## Performance

- **Batch TET fetch:** replaced the per-reference `GET /by_reference/{curie}`
  fan-out with a single `POST /topic_entity_tag/by_references` (chunked), with a
  per-reference fallback if the endpoint isn't available.
- **Reuse search biblio:** the grid now reuses the biblio the search already
  returned, skipping the per-row `/reference/{curie}` fetch for AGRKB curies.
- Net: a typical search page drops from ~2×N requests to ~1–2.

## Testing
- Added `fetchTetsBatch` tests (success, dedupe, empty input, per-reference
  fallback). Existing `resolveOne`/`fetchTets` tests unchanged.

### DIFF
--- a/src/actions/searchActions.js
+++ b/src/actions/searchActions.js
@@ -160,7 +160,10 @@ const getSearchParams = (state) => {
 
   // Negated TET facets. Source method / source evidence assertion exclusions use whole-reference
   // semantics (drop a reference if ANY of its tags matches) and apply in both single- and
-  // multi-tag modes. Confidence level negation keeps its existing single-tag-only behavior.
+  // multi-tag modes. Confidence level negation applies in both modes too: the backend attaches
+  // it as a must_not inside each positive nested tag query (add_nested_query), so it correctly
+  // excludes NEG tags per-tag whether one or many TET facets are selected (it just needs at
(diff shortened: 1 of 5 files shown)


sweng66:
@claude Can you check again?

azurebrd (approved):

sweng66:
Thanks @azurebrd

PR #611, opened 2026-06-25
add reusable entity-counts-by-MOD summary and store-backed TET loading

sweng66:
Add EntityCountsByMod, a reusable component that shows the unique entity count by entity type and MOD (from secondary data provider) for a reference, with no distinction between manual and automated tags. It lists only MODs with associated entities, only entity types with count > 0, sorted alphabetically. Placed in the workflow and topic/entity editors and kept self-contained for future reuse (biblio display, etc.).

Load topic entity tags into the redux store on any reference-based (Biblio) page so the summary and other consumers can reuse them. Refactor TopicEntityTable to consume the store instead of fetching its own copy, applying display transforms via a memo without mutating redux state, and force-refreshing only after an add/edit completes.

### DIFF
--- a/src/actions/biblioActions.js
+++ b/src/actions/biblioActions.js
@@ -976,6 +976,68 @@ export const fetchReferenceFiles = (referenceCurie, forceRefresh = false) => {
   };
 };
 
+export const setTopicEntityTags = (tags, referenceCurie) => {
+  return {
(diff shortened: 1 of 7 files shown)


sweng66:
@claude Check again?

sweng66:
@claude Check again?

valearna (approved):

sweng66:
Thanks @valearna

PR #612, opened 2026-06-29
Speed up Search topic grid display 

sweng66:
Updated the UI batch hook to consume API-provided entries, counts, and filtered tags.

Updated the topic grid renderers, filters, and sorting helpers to use the API aggregates instead of rebuilding those buckets repeatedly in the browser.

Preserved fallback behavior for older API responses by locally grouping raw tags only when API entries are absent.

Kept raw TETs available for curator validation actions/status, so validation behavior should remain unchanged.

Forwarded entity filters from search state to the topic grid batch request when available.


sweng66:
@claude Can you check again?

sweng66:
@claude Can you check again?

valearna (approved):

sweng66:
Thanks @valearna

PR #613, opened 2026-06-30
 consume server validation aggregates and validate write path

sweng66:
Wire the TET validation grid to the server-side aggregates so the Validation column renders, sorts and filters without re-deriving from raw tags, and route curator validations through the new thin write endpoint.

- ValidationCell renders from the server `validation` cell (by_curator / positives / negatives), falling back to deriving from raw tets when absent.
- groupTets validationState/cellPredicate prefer the per-cell server validation / filter_flags; "my validation present" stays raw-tag based since the server omits it (needs curator identity).
- innerColumnUtils VALIDATION sort/filter pass the whole cell value so the server-aggregated state/rank is used.
- TetValidationGrid plumbs validation/filterFlags into each cell value and, after a validation, applies the single recomputed cell in place (applyValidatedCell) instead of re-fetching/re-aggregating the batch; falls back to a per-reference refetch when the row is not server-aggregated.
- CellValidationStrip writes via POST /topic_entity_tag/validate (server resolves the curator source) and hands back the recomputed cell.

Also fix a pre-existing useReferenceTets test that expected entries {} where the code returns null for a response with no entries map.

### DIFF
--- a/src/components/refs_tet_validation/TetValidationGrid.jsx
+++ b/src/components/refs_tet_validation/TetValidationGrid.jsx
@@ -49,6 +49,16 @@ import {
 import { debug } from './helpers/debug';
 import './TetValidationGrid.css';
 
+// Per-cell filter flags for a topic that is a column but carries no tags on a
+// given (server-aggregated) reference — the server omits such topics from its
+// filter_flags map, so "absent" means all-false.
+const EMPTY_CELL_FILTER_FLAGS = Object.freeze({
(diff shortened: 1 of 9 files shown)


sweng66:
Can you check again? @claude

sweng66:
@claude Check again?

valearna (approved):

sweng66:
Thanks @valearna

PR #616, opened 2026-07-10
image facets/icon and support all XREF search (including PDD/GEO IDs)

sweng66:
Surface the reference image data now provided by the search API:
- New "Images" facet category with "Image permission" (can_display_image) and "Has images" (has_image) facets, so curators can filter to references with image permission and/or figures already uploaded.
- Show a green image icon on a result when image_count > 0, with a tooltip of the figure count.

Also stop hiding dataset cross-references (PDB, GEO) in search results so curators can see the xref that matched their search.


azurebrd (approved):
Thanks @sweng66

sweng66:
Thanks @azurebrd

PR #619, opened 2026-07-13
fix(scrum-6295): show nXML- and TEI-derived markdown, hide raw tei files

valearna:
## Summary

[SCRUM-6295](https://agr-jira.atlassian.net/browse/SCRUM-6295): the Biblio file-management table only displayed `_merged` markdown — markdown converted from nXML (`_nxml`) and TEI (`_tei`) sources was never shown. Raw `tei` files are also deprecated and should no longer be shown.

- **Root cause for `_nxml`**: nXML-derived markdown (`file_class: converted_merged_main`, display_name `<source>_nxml`) groups under the nXML source row, but `getConvertedFileClasses('nXML')` returned `[]`, so the nXML row's Converted Files column stayed empty. nXML rows are now treated like main rows.
- **Raw tei files** are no longer rendered as links. `'tei'` stays in `ALL_CONVERTED_FILE_CLASSES` so tei rows remain filtered out of the file table.
- **TEI-derived markdown** (`_tei` suffix) still displays on the main row (TEI files share the main PDF's display_name, so the stripped base matches the main row).
- **Dedup**: main PDFs and nXML files usually share a display_name (~180k refs on dev DB), which would have shown the same conversion on both rows. `_nxml` markdown is pinned to the nXML row; other conversions to main/supplement rows.
- The grouping/selection logic is hoisted to exported pure helpers with unit tests (15 cases).

Verified against dev-DB data: all 290,140 `converted_merged_main` rows end in `_tei`/`_nxml`/`_merged` and every one now anchors to a visible row, except ~309 duplicate-conversion artifacts (refs with two TEI files) that stay hidden as redundant copies.

## Test plan

- `CI=true npx react-app-rewired test src/components/biblio --watchAll=false` — 31 tests pass
- `npx eslint` on changed files — clean
- Traced real refs through the logic: `AGRKB:101000001212217` (markdown now visible on the nXML row) and the `PMC7971549` ref (tei-derived markdown on the main row, no raw tei link)

[SCRUM-6295]: https://agr-jira.atlassian.net/browse/SCRUM-6295?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/biblio/BiblioFileManagement.js
+++ b/src/components/biblio/BiblioFileManagement.js
@@ -820,7 +820,9 @@ const CONVERTED_FILE_CLASSES = {
   ]
 };
 
-// All converted file classes (for filtering)
+// All converted file classes (for filtering). 'tei' stays listed so tei files
+// are kept out of the file table, even though tei links are no longer rendered
+// in the Converted Files column.
 const ALL_CONVERTED_FILE_CLASSES = [
   ...CONVERTED_FILE_CLASSES.main,
   ...CONVERTED_FILE_CLASSES.supplement,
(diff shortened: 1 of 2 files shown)


valearna:
@claude Re: the map-collision concern (`PMC789_merged` + `PMC789_nxml` both `converted_merged_main` on the same base) — I investigated this against the backend and the dev database. The scenario is structurally unreachable, on three independent grounds:

1. **Batch pipeline**: `get_pdfs_missing_xml` in `pdf2md_utils.py` only selects references with *no* nXML and *no* tei files, so PDFX `_merged` outputs are confined to references where `_nxml` markdown can't exist.
2. **On-demand API path**: `pending_main_sources` (always called with `prefer_nxml=True`) returns early when an nXML source exists — quoting the backend comment: "nXML present (converted or pending) — by convention we don't ALSO convert main PDFs when nXML is the chosen source for this reference." The nXML conversion is the canonical main markdown by design.
3. **Data**: a dev-DB query for bases with more than one `converted_merged_main` markdown per reference returns **0** across all 290,140 rows.

The single-file-per-class map deliberately relies on this backend one-canonical-main-source invariant, so no UI-side hardening was added. If that convention ever changes, the map's inner key would need to include the derivation suffix — but that's the backend change's concern to carry, not a latent UI bug today.

sweng66 (approved):
Thanks @valearna

valearna:
thanks @sweng66 !

PR #620, opened 2026-07-13
 label Images facet breadcrumbs with facet name and Yes/No instead of true/false

sweng66:
…o instead of true/false


azurebrd (approved):

sweng66:
Thanks @azurebrd

PR #621, opened 2026-07-15
show first pass curation facet in Workflow Tags + hide Edit/Delete for non-ABC tags 

sweng66:
Register the first_pass_curation Workflow facet returned by the API as a new parent under the "Workflow Tags" group (label + group membership + render whitelist so it is queried without a spurious .keyword suffix).


sweng66:
@claude Can you check again?

valearna (approved):

sweng66:
Thanks @valearna

PR #623, opened 2026-07-17
SCRUM-5781: show ml_model embedding columns in the Models table

valearna:
## Summary
Surface the new `ml_model` ABC-embedding columns in the Models report grid so curators can see which models were trained on the ABC's precomputed embeddings.

Adds two columns to `ModelsTable.js` (both to the column defs and the Hide/Show list):
- **Embedding Profile** (`embedding_profile`) — the ABC embedding profile; blank/NULL for legacy BioWordVec models.
- **Embedding Version** (`embedding_version`).

Both are populated from `/ml_model/all` (the fields are added in `agr_literature_service` PR #1245).

## Depends on
`agr_literature_service` PR #1245 — the API must return `embedding_profile`/`embedding_version` for these columns to populate. Until that's deployed the columns simply render blank.


sweng66 (approved):
Thanks @valearna

PR #627, opened 2026-07-24
Scrum 6272

azurebrd:


azurebrd:
@claude review again, and note Claude Code says 

  Finding 1 — premise doesn't hold ❌

  by_strain_designation does not 404 on no match; it returns 200 []. So for a "smith" name search, by_strain_designation resolves with an empty array,
  Promise.all resolves, and the name matches display. The described failure (name search broken by a strain 404) can't occur. The status === 404
  handling in Create.js that was cited is defensive/dead code for these endpoints — I wrote it belt-and-suspenders; it doesn't prove they 404.
  Residual point: Promise.all would still drop both results if one endpoint hard-errors (500/network). But that's not the reviewed scenario, both
  endpoints hit the same service (so a one-sided 5xx is unlikely), and abort is already handled correctly by the hook's seq/signal.aborted guard. I
  don't think a change is warranted; I can switch to allSettled purely as optional hardening if you want.

  Finding 2 — premise doesn't hold ❌

  response_model=List[...] means FastAPI always serializes a JSON array, even for a single exact strain match ([lab], never a bare {...}). So
  Array.isArray(byStrain.data) is always true and nothing is dropped. The single-object handling in Laboratory.js/Create.js exists because those
  blocks also normalize the curie/id/xref endpoints (/laboratory/{id}, by_laboratory_cross_reference), which do return a lone object — LabCuriePicker
  never calls those. So its array-only check is correct and complete.

  Finding 3 — real ✅

  Confirmed by tracing: loading a field-less free-text URL (/lab?q=unc-13) pushes ?q=unc-13&field=name, and pressing Back re-triggers the effect
  (currentFieldRef='name' !== validField=null), which re-pushes — a Back trap plus a redundant fetch. Curie/xref links are immune (no field appended).
  Fixing now with history.replace on the URL-sync path.

azurebrd:
@claude review

azurebrd:
@claude review

azurebrd:
@claude review

azurebrd:
@claude review

azurebrd:
@claude review

azurebrd:
@claude review

valearna (approved):

azurebrd:
Thanks @valearna !

PR #630, opened 2026-07-27
Fixed Search query builder to only allow a tag to carry each field at most once

sweng66:
Fix code so a tag can carry each field at most once (a tag can't be topic=A AND topic=B), so disable already-used field types in each row's dropdown, default new rows to the first unused type, and disable "Add field" once every type is present. This keeps the interface and compiled query in agreement.


sweng66:
@claude Can you check again?

sweng66:
@claude Can you check again?

valearna (approved):

sweng66:
Thanks @valearna

PR #631, opened 2026-08-03
Scrum 6311

azurebrd:


azurebrd:
@claude review again
● All four review fixes done and verified — hook/util tests 11 pass (added the bad-regex test), eslint clean, build compiles:

  1. Dirty-forever mismatch → labPositionValue/relationshipValue now keep the id as a string in state; the select value, change events, and saved-key
  snapshot all compare string-vs-string, and bodyFn converts to Number only when sending. No more permanent "dirty" / redundant re-PATCH.
  2. Fetch-failure lockout → both loadVocabulary/loadCheckPatterns now cache.delete() on rejection, so a transient 401/500 no longer poisons the
  module cache for the whole session — the next mount retries.
  3. Unguarded new RegExp → regexFor wraps the compile in try/catch and returns null (treated as "no pattern") for Python-only syntax, so one bad
  backend pattern can't crash the editor. Covered by a new test.
  4. Loading-window flash → the injected current-term option only gets the (obsolete) label + disabled once the vocab has loaded (loading gate), so a
  valid term isn't briefly shown as "N (obsolete)".

  On the Minor (dismissing a format error does nothing): I left it deliberately — curieError is a live validation state, so it should persist until
  the curie is fixed; the × only clears server (_error) messages. Happy to suppress the dismiss affordance for pure format errors if you'd rather.

  I also didn't touch the Number() coercion beyond the string-normalization, since lab_position/relationship are always table-backed ids (never
  string-valued vocabs) — the static vocabs send their string value unchanged and are never Number()'d.

azurebrd:
@claude review again

sweng66 (approved):
Thanks @azurebrd

azurebrd:
Thanks @sweng66 !

PR #632, opened 2026-08-05
feat(scrum-6330): align topic grid column names and values with the rest of the UI

valearna:
## Summary

Closes [SCRUM-6330](https://agr-jira.atlassian.net/browse/SCRUM-6330).

The Topic grid's "Assessment by Biocurator" column is derived from the same manual-assessment ECO code as the "data" / `has_data` column, but rendered its values as `positive` / `negative` while that column — and the advanced topic & entity search facets now built on it — uses Y/N. Nothing in the UI told curators the two were the same thing.

Per the ticket's decisions:

| Where | Before | After |
|---|---|---|
| Column header | `Assessment by Biocurator` | `Data assessment by biocurator` |
| Cell badge | `positive` / `negative` | `Y` / `N` |
| Cell badge | `assessment conflict` | `conflict` |
| Filter checkboxes | `unvalidated` / `positive` / `negative` | `unassessed` / `Y` / `N` |
| Filter popup title | `Assessment status` | `Data assessment by biocurator` |

The source-side columns are now collected under one nested group header rather than renaming a single column:

```
┌──────────────────────────────────────────────────────────────────┐
│                        Topic name (ATP:…)                        │
├──────────────┬───────────────────────────────────────────────────┤
│              │        Data assessment by other sources           │
│              ├─────────┬──────┬─────────┬──────────┬─────────────┤
│ Data         │ Sources │ Data │ conf sc │ conf lvl │ note        │
│ assessment   ├─────────┼──────┼─────────┼──────────┼─────────────┤
│ by biocurator│textpresso│  Y  │  0.91   │   high   │     📝      │
│    [ Y ]     │ abc_ent… │ {3}E │ .40–.88 │    2     │             │
└──────────────┴─────────┴──────┴─────────┴──────────┴─────────────┘
```

All five entry-driven columns went under the group, not just Sources and Data: `buildEntries` skips curator-submitted tags outright (`helpers/buildEntries.js:112`), so every one of them describes exclusively non-biocurator sources and they iterate the same per-source rows in lockstep. Only the biocurator column sits outside. `Sources` and its tooltip therefore revert to their original wording — the group header carries the phrase now.

Header tooltips were rewritten to state that Y/N here mean the same as the `has_data` facet in advanced search.

## Labels are display-only

`positive` / `negative` / `conflict` are the server's `validation.state` aggregate values **and** what gets persisted in saved AgGrid filter models. Renaming them would silently invalidate both, so the keys are untouched and the wording comes from a `VALIDATION_STATE_LABELS` map in `helpers/groupTets.js`, threaded into the filter through a new display-only `valueLabels` param. A test pins that invariant.

Side effect worth flagging: `unvalidated` now displays as `unassessed` in the live filter. That wording already existed in the unused `ValidationFilter.jsx`, which now reads from the shared map so the two filters can't drift.

## Layout fixes the longer headers forced

- `makeInnerColumn` was silently dropping the `wrapHeaderText` / `autoHeaderHeight` flags the validation column already passed — harmless with short labels, but a 32-character header would have been clipped. Now forwarded for all inner columns.
- AgGrid's autosize measures headers unwrapped on a single line, so the validation column is capped at `maxWidth: 150`; without it every topic group would widen ~35px just to fit the header phrase.
- New `.tetv-topic-subgroup-header` tone for the nested band, plus a rule for `ag-header-group-cell-no-group` (AgGrid's filler cell above the biocurator column) so it stays blank instead of showing a stray band.

Leaf `colId`s are all unchanged, so `allowedInnerColumnIds`, the filter/sort pruning on topic visibility changes, the autosize `__`-suffix heuristic, and the sticky scrollbar's header-height `ResizeObserver` are unaffected by the regrouping.

## Test plan

- `npx eslint` clean on all changed files.
- `CI=true npx react-app-rewired test` — 176 tests pass across 16 suites, including 6 new `ValidationCell` tests covering Y, N, conflict, the unassessed-strip fallback, and the filter-key invariant.
- `src/App.test.js` still fails on a pre-existing d3 import error in `WorkflowDiagram`, confirmed identical on unmodified `main`.
- **Not verified:** no suite renders the grid's column definitions, and the rendered three-row header has not been looked at in a browser. The header height, the wrapped `Data assessment by biocurator` label at its 150px cap, and the blank filler cell above it are worth a visual check before merge.

[SCRUM-6330]: https://agr-jira.atlassian.net/browse/SCRUM-6330?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/refs_tet_validation/TetValidationGrid.css
+++ b/src/components/refs_tet_validation/TetValidationGrid.css
@@ -353,6 +353,25 @@
   color: #475467;
   font-weight: 600;
 }
+/* Nested group inside a topic group: the "Data assessment by other sources"
+   band that spans the source label, its Y/N/{N}E tag and the optional
+   confidence / note columns. Toned between the topic-name band above it and
+   the leaf sub-headers below so the three levels stay distinguishable. */
+.ag-header-group-cell.tetv-topic-subgroup-header {
+  background: #eef3f7;
+  font-size: var(--tetv-secondary-font-size);
+  font-weight: 700;
+  color: #3d5566;
+  border-right: 1px solid #d0dce7;
+  border-bottom: 1px solid #dbe4ec;
+}
+/* The biocurator column has no nested group above it, so AgGrid renders an
+   empty group cell in that row. Keep it blank rather than showing a stray
+   band of the sub-group tone. */
+.tetv-grid-wrapper .ag-header-group-cell-no-group {
+  background: #f8fafc;
+  border-right: 1px solid #d0dce7;
+}
 /* Alternating row tones for readability. Each paper is separated by a
    prominent slate-grey divider so curators clearly see where one paper ends
    and the next begins. */
@@ -383,6 +402,9 @@
   font-size: 1rem;
   font-weight: 700;
   padding: 7px 14px;
+  /* Y and N are single characters — keep both pills the same size so a column
+     of assessments reads as one consistent shape. */
+  min-width: 2.75rem;
   border-radius: 999px;
   text-align: center;
   white-space: normal;
--- a/src/components/refs_tet_validation/TetValidationGrid.jsx
+++ b/src/components/refs_tet_validation/TetValidationGrid.jsx
@@ -17,6 +17,7 @@ import {
   groupTetsByTopicAndSource,
   sourceLabel,
   normalizeCurie,
+  VALIDATION_STATE_LABELS,
 } from './helpers/groupTets';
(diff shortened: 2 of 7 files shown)


sweng66 (approved):
Thanks @valearna

valearna:
thanks @sweng66 !

